PR #56, opened 2015-03-06
Adding tooltip to buttons

fclaussen:
I'm sorry, I was kinda lost in your code. You have some files inside so many folders and using cdn code that I had to change. I mean, this is my first time collaborating and It all got a little bit out of control.

However, if you check what I've done, you will see where I am going with this. My tooltip code works, you just need to use it accordingly since I altered the .css file directly.

**What I've done**
* I altered the index file to use local css and js so I could see changes.
* Added "data-tooltip" to buttons
* Added css rules directly in the .css file (I could not find the existing rule I was looking for in the sass file), to style and position the toopltip. Since I'm clearly not a designer, you just need to set the colors to your taste.

Hope you enjoy it and you can take something from it.


SamPotts:
Hey @fclaussen, 

Thanks for that. Yes, the reason it's broken down into folders is the standard build process with projects like these. You tend to have a `src` folder where the source files live (uncompressed JS, LESS, SASS, etc) and are then built using build systems like gulp or grunt which output compressed, production ready files to `dist`. If you're not familiar with this setup, I'd encourage you to have a play. Once you're learnt it, you'll wonder how you worked before! :) 

As I said in the original issue, my preference would be to make it an option (not everyone may want them) and use the original `.sr-only` spans and labels for the tooltip so there's not doubled markup but I've not had a deep dive into your solution yet. 

I'll take a look at what you've done and go from there. :+1:

fclaussen:
Ok. Great. I'll try to play around with this file template.

PR #57, opened 2015-03-06
add tooltip functionality

fclaussen:
I'm sorry. This is my first time collaborating to github and I got a little lost in your file structure. But you can see that my code works.

**What was done**
* I changed index.html to use local css and js so I could see changes.
* I added some rules to the .css file since I did not find the rule I was looking for under the scss file. The class I was looking for was .player-controls-playback.
* I changed the minified ,js file that renders the controls and added data-tooltip to it.

It might not look beautiful, since I'm not a designer and I believe that this is something only you can decide how it needs to be, however, it works!

Hope you like it and at least you can take something out of it.


josisusan:
https://github.com/Selz/plyr/issues/50

SamPotts:
Hey @fclaussen , I've done it a slightly different way but thanks for your work on this :+1:

PR #70, opened 2015-03-30
Adds the full duration time to the player

franks921:
Hi @SamPotts 

Here is a screenshot of what this looks like: 

Sample

I did not go into too much details, as I am not sure if the design is according to your vision.

I am happy to do the extra work on this PR to get it to your standard, if this is a feature you would like to include?

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -43,7 +43,8 @@
                 played:         ".player-progress-played"
             },
             captions:           ".player-captions",
-            duration:           ".player-duration"
+            duration:           ".player-duration",
+            fullDuration:       ".player-full-duration"
         },
         classes: {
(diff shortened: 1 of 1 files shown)


SamPotts:
Hey @franks921, funny that you added this as I just this minute added displaying duration at startup as an option. I'm going to make some changes to how the controls work soon though so I'll use what you've done for that. I was going to add an option for hiding/showing some controls etc so this will be useful for that. It'll probably be v1.1 or something. Then I need to crack on with playlists... 

Thanks for your work :+1: Looks the same as how I've done it.

franks921:
Cool stuff @SamPotts. Please let me know if there is anything I can help with :+1:

PR #76, opened 2015-04-10
Make SCSS variables optional

robinpoort:
When I was using this great plugin a couple of weeks ago I stumbled upon a very small but irritating little 'bug'. I am compiling my own CSS file consisting out of multiple plugins and custom SCSS. When compiling and using the plyr.scss source file it's impossible to overwrite the variables without editing the file itself. I needed to adjust the main blue color to green to match the player with the rest of the site. I'm currently `@import`ing the plyr.scss file and overwriting the color in a separate file that I'm loading after the plyr.scss file.

But of course it would be better to be able to overwrite the variables beforehand. So I added `!default` to each variable making it overwritable before the `@import` like so:

```
$blue: #81C3B3;
@import "plyr";
```

Additionally we could also make some of the `classes` and `@mixin`'s optional. So if your project already has a `clearfix` mixin you don't need it again in the plyr.scss file. Same goes for the 'sr-only` class; if you already have it then you don't really need it again since it gets you duplicate selectors. Also the tab-focus and font-smoothing might be two things you want to handle yourself to match the rest of your page. But that should probably be another issue.

### DIFF
--- a/src/sass/plyr.scss
+++ b/src/sass/plyr.scss
@@ -5,50 +5,50 @@
 // Variables
 // -------------------------------
 // Colors
-$blue:                    #3498DB;
-$gray-dark:               #343f4a;
-$gray:                    #565d64;
-$gray-light:              #cbd0d3;
-$off-white:               #d6dadd;
+$blue:                    #3498DB !default;
+$gray-dark:               #343f4a !default;
+$gray:                    #565d64 !default;
+$gray-light:              #cbd0d3 !default;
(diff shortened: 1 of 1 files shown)


SamPotts:
Sounds good to me. I've been meaning to clean up the LESS and SASS/SCSS a bit so I'll take a look at your other suggestions. Cheers :+1:

robinpoort:
:+1:  If you need any help with that just let me know!

PR #111, opened 2015-08-13
Formatting & Optimization

calvintam236:


SamPotts:
Thanks for doing the housekeeping stuff. 

The only change that concerns me is now you've broken the JSHinting by changing the quotations to mixed for the HTML commit. Maybe revert that or just change everything to be single if you really want to change it and then I'll merge it. 

I'd suggest using and editor that supports JSHint. Sublime has JSHInt Gutter which is good.

calvintam236:
I just tested the RAW code on http://jshint.com. I do not receive any errors..

SamPotts:
screen shot 2015-08-13 at 5 54 04 pm

> Mixed double and single quotes

SamPotts:
What was your reasoning behind that change anyway? The others seem fine.

calvintam236:
Fixed. As a web developer, most of us use double quotes in html coding. If you validate the source, it will give us warning of "Mixed double and single quotes."

SamPotts:
Heh, "us" includes me. I'm also a full time designer and developer and I'm aware that both single and double quotes work fine for both HTML and JavaScript. It's merely a personal preference thing. Happy to go with singles for this though if it's needed. HTML in JavaScript is always going to be a trade off. Having HTML in JavaScript is a bit of an anti-pattern in itself. Templates should be used (like in the docs example). 

Also, your "fix" for equal signs isn't really a fix. Nothing was broken there. Are you aware of the difference between `==` and `===` ? I only use `===` (which also checks type) when it's really needed. For example 'typeof` will always return a string so `===` is not really necessary. Other cases such as checking for `null` or `0`, it's recommended. 

I'll merge this in a bit later today if I don't find anything else.

calvintam236:
I looked at your source code, and most typeof cases use ```===```, not ```==```. I also checked https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof, and all the examples use ```===```. That's why I "fixed" them.

SamPotts:
All live in v1.3.4 :+1:

PR #126, opened 2015-12-03
cancel open requests when destroy is called

oemueller:
When destroy is called I expected the media to stop playing AND to stop loading data from the server. This is also important if you create multiple instances of a video player from time to time like on a one-pager website.

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -1879,6 +1879,12 @@
             // Clone the media element to remove listeners
             // http://stackoverflow.com/questions/19469881/javascript-remove-all-event-listeners-of-specific-type
             var clone = player.media.cloneNode(true);
+
+            // Cancel open requests
+            player.media.pause();
+            player.media.setAttribute('src','');
+            player.media.load();
+
             player.media.parentNode.replaceChild(clone, player.media);
         }
 


oemueller:
Hi, this fix solved a serious problem for me. Any reason not to merge it ? :-)

SamPotts:
I'm just unsure why we need to kill the sources when destroying the plyr instance. The point of `.destroy()` is to revert back to pre-initialisation state. Removing the source might be a bit heavy handed. I think we should make it an option you can pass to the method (boolean) and add it to the docs as an option to remove sources. Anyway, to remove all sources, it'd be better to call `_removeSources()` anyway as it will remove child `<source>` elements.

oemueller:
the source is removed after cloning the node, so the original sources should not be affected. (I try to test this) It's just about stopping the request on the media resource by loading a "null" resource. 
Without this I was not able to create like more then 6-7 plyr instances one after another. 
....
Another option would be to create a method "unload" or "stop" to get rid of open requests.

SamPotts:
Sorry I never merged this but I was working on the new API method at the time. I've used this code in a fix I just pushed though, thanks :+1: 

https://github.com/Selz/plyr/issues/174

PR #134, opened 2016-01-10
Fixed embedded Plyr not updating plyr.media.paused

gurupras:
Added code to properly update plyr.media.paused in embedded players so that it correctly reflects the current video state.


SamPotts:
Cheers for this :+1:

PR #139, opened 2016-01-13
Updated API function 'plyr.source()' to get media source

gurupras:
Currently, there is no simple way of acquiring media source.
This commit updates the 'source' API function to allow developers
to acquire the video source regardless of whether the video
is a youtube-embed or HTML5 video.

Future embeds just need their respective URL fetching code added
to this common function.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -356,10 +356,10 @@ Here's a list of the methods supported:
   </tr>
   <tr>
     <td><code>source(...)</code></td>
-    <td>String or Array</td>
+    <td>String or Array or (null|undefined)</td>
     <td>
-      Set the media source.
-      <br><br> 
+      Get/Set the media source.
+      <br><br>
       <strong>string</strong><br>
       <code>.source("/path/to/video.mp4")</code><br>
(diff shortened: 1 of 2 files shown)


gurupras:
This should allow developers a common way of pulling source. 
This is especially helpful when a page has multiple video elements of both types.

SamPotts:
Cheers for this :+1: 

- That's not going to work for those that use the `src` attribute on the `video` or audio element itself
- I'd probably return all the child source elements as an array of objects, with keys `src` and `type`
- You don't need the semi colon at the end of a function definition. 
- The comments standard we're using is `//`
- Missing the curly brackets for the first if statement

Happy to make these changes if you don't have time.

gurupras:
I don't fully understand points 1 and 2.
Is it OK if I leave those to you?

I can definitely do 3, 4, 5. I will create a new pull request with 3, 4, 5 done.

SamPotts:
Awesome. Yeah no worries on 1 and 2. I can take care of them.

PR #140, opened 2016-01-13
Updated API function 'plyr.source()' to get media source

gurupras:
Currently, there is no simple way of acquiring media source.
This commit updates the 'source' API function to allow developers
to acquire the video source regardless of whether the video
is a youtube-embed or HTML5 video.

Future embeds just need their respective URL fetching code added
to this common function.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -356,10 +356,10 @@ Here's a list of the methods supported:
   </tr>
   <tr>
     <td><code>source(...)</code></td>
-    <td>String or Array</td>
+    <td>String or Array or (null|undefined)</td>
     <td>
-      Set the media source.
-      <br><br> 
+      Get/Set the media source.
+      <br><br>
       <strong>string</strong><br>
       <code>.source("/path/to/video.mp4")</code><br>
       This will set the <code>src</code> attribute on the <code>video</code> or <code>audio</code> element.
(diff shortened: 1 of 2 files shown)


gurupras:
Are you suggesting that we should use `player.getElementsByTagName('source')` in case of native video and then return all the sources by type?

gurupras:
In case of native media, can we use `player.media.currentSrc`?
This seems to work regardless of whether there is a `source` element below `media` or `media.src` is set directly.
I tested this out a little bit. You can find the fiddle [here] (https://jsfiddle.net/gurupras/kp55x8k5/)

SamPotts:
That would work yep.

gurupras:
I've changed the logic for local videos to use `player.media.currentSrc`

SamPotts:
Nice one fella :+1:

gurupras:
Quick question, would you rather I work off of master / develop?
Just asking so it'll be less hassle for you to merge.

SamPotts:
Probably develop. I'm working on v1.5 in that branch (Vimeo support added) and a few other fixes:
https://github.com/Selz/plyr/milestones/1.5

SamPotts:
One more thing. Setup jshint in your editor and it'll show formatting errors. Just realised you used double quotes and we changed to single recently. Not a big issue as I'll manually merge and fix it in develop now.

SamPotts:
This is merged to Develop now with the changes I mentioned above.

PR #141, opened 2016-01-14
Added new configuration option 'handlers'

gurupras:
This option allows developers to attach their custom handlers
before internal plyr's handlers. This should enable developers
to have full control over all the buttons including the ability
to stop event propagation if desired.


gurupras:
Given how difficult (impossible?) it is to change the order of event handlers once they are bound, and given how plyr's event handlers are all registered early ( `setup()` ), this commit should enable developers to have full control over all the controls.

src/js/plyr.js
```diff
@@ -2170,34 +2182,43 @@
...
 
             // Seek
-            _on(plyr.buttons.seek, inputEvent, _seek);
+            _registerHandler(plyr.buttons.seek, inputEvent, config.handlers.seek, _seek);
 
             // Set volume
-            _on(plyr.volume, inputEvent, function() {
-                _setVolume(this.value);
+            _registerHandler(plyr.volume, inputEvent, config.handlers.volume, function() {
+                _setVolume(plyr.volume.value);
```
SamPotts: This should be `this.value` otherwise it won't get the value of the input?

src/js/plyr.js
```diff
@@ -2170,34 +2182,43 @@
                 });
             }
 
+            function _registerHandler(element, event, userHandler, defaultHandler) {
+                _on(element, event, function(e) {
```
SamPotts: To be consistent, could you use the full `event` variable name so it's clear.

src/js/plyr.js
```diff
@@ -68,6 +68,18 @@
...
+            seek:           null,
+            play:           null,
+            pause:          null,
+            restart:        null,
+            rewind:         null,
+            forward:        null,
+            mute:           null,
+            volume:         null,
+            captions:       null,
+            fullscreen:     null
```
SamPotts: Could you make sure the tabbing matches the rest of the code? I can't remember if we use 4 spaces or a hard tab but can you just make sure it's consistent, cheers

SamPotts:
My only concern is the the passing of `this` will be broken. I'll merge it though and take a look.

src/js/plyr.js
gurupras: I tested this out extensively and it works. Using `this.value` throws an error stating `TypeError: 'this' is undefined`. Also, I'm probably wrong, but I believe plyr.volume *is* the input?

src/js/plyr.js
gurupras: Ah crap. I didn't check for this. Sorry.

gurupras:
I believe I've incorporated all of your comments. Let me know if something's still broken and I'll work on fixing it. For now, I've changed the variable name to `proxyEvent` as you suggested.

SamPotts:
It's cool. All minor and sorted. I just moved the `_registerHandler` up to where `_on` and `_off` etc are and renamed to `_proxyHandler` as it's kinda like a middle man. You were right about the `this` references but what I've done is:

```
userHandler.apply(element, [event]);
```

That will send the element the handler is bound to as `this` to the callbacks which I thought might be useful for those wanting to add their own handlers.

Cheers for that :+1: will add to the v1.5 changelog.

gurupras:
Not sure how I didn't notice this earlier, but the current implementation has some unintended consequences.

For example, Using the early listener and `preventDefault()`/`stopImmediatePropagation()`, I should ideally be able to prevent the `defaultHandler` from executing. However, the current implementation only ensures that the `userHandler` runs before `defaultHandler`.

An example of this can be found in this [fiddle](http://jsfiddle.net/gurupras/fphym29o/)
I'll try and quickly come up with a PR for this.

PR #142, opened 2016-01-14
Added API call to update track element

gurupras:
Minor change to add API call to enable modifying the track element.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -469,6 +469,21 @@ Here's a list of the methods supported:
       Returns the current media source url. Works for both native videos and embeds.
     </td>
   </tr>
+    <td><code>track(...)</code></td>
+    <td>Object or Array</td>
+    <td>
+      Get/Set the track source.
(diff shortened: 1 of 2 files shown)


SamPotts:
Will this just append new `track` elements or should we remove old ones?

gurupras:
Currently, this just appends track elements. This should allow dynamic loading of track elements.
My reasoning for now handling delete is this:
When the video changes (`_updateSource()` called), it makes no sense to preserve track elements. So I would think that this is the right place to get rid of existing track elements. I also can't think of any (other) cases where we may want to get rid of loaded track elements.

In terms of full control, it makes sense to give them an API call to delete track elements, but I'm not sure how often it will be used.

Given my lack of knowledge on this, I will leave the decision up to you. Let me know if you want me to make changes and I will gladly get them done.

gurupras:
closing since it's probably irrelevant now.

PR #150, opened 2016-01-25
Remove global preinstall script

marvinhagemeister:
There are a lot of systems where the user has no rights to install packages globally. It isn't needed for gulp anyway, since it's installed locally anyway.


SamPotts:
Cheers for this :+1:

PR #151, opened 2016-01-26
Fix broken npm require/import

marvinhagemeister:
The `main` entry inside `package.json` must always point to the main source file, otherwise the package can't be imported.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "1.5.9",
   "description": "A simple, accessible and customizable HTML5, YouTube and Vimeo media player",
   "homepage": "http://plyr.io",
-  "main": "gulpfile.js",
+  "main": "src/js/plyr.js",
   "dependencies": {},
   "devDependencies": {
     "gulp": "^3.9.0",


SamPotts:
Thanks. The NPM stuff is new to me :)

marvinhagemeister:
No worries, we just switched to npm to manage frontend dependencies like a week ago at work. We're a huge fan of your library here! Any chance you could publish a new release to npm in the coming days?

SamPotts:
Awesome, glad you like it. 

I did this today? Is this what you mean? 
https://github.com/Selz/plyr/issues/146

Should have the latest version on there now. I've also updated the readme but I can push your change to package.json if you like?

SamPotts:
Just updated it to v1.5.10 with your changes in :+1: 

https://www.npmjs.com/package/plyr

marvinhagemeister:
You right, that it was already available. But without the modified `main`-entry it's impossible to import plyr. Thanks for making a quick v.1.5.10. Your awesome!

PR #152, opened 2016-01-26
Swap CommonJS and AMD module check so that CommonJS is checked first

marvinhagemeister:


SamPotts:
Thanks again, sorry again UMD etc is new to me.

PR #154, opened 2016-01-27
Get the proper reference to window on initialization.

marvinhagemeister:
The previous code assumed that `this` points to the browsers
window object, which is not the case when using a module bundler.

So we check first if `window` has been defined by our bundler before
falling back to `this`.

(taken from jQuery's codebase)


SamPotts:
Thanks :+1:

Issue #158, opened 2016-02-04
Set default colors only if they are null (SASS)


ivanvermeyen:
Hi

I was trying to set some colors for the controls, but it seems you are overwriting the vars `$plyr-control-color` and `$plyr-control-color-hover`, based on the lightness of `$plyr-controls-bg` and `$plyr-control-bg-hover`.

https://github.com/Selz/plyr/blob/master/src/sass/plyr.scss#L26-L45

I could change the 'gray' vars, but `$gray: red;` doesn't feel that good :)

Maybe you could add a check to see if `$plyr-control-color` and `$plyr-control-color-hover` are null before overwriting them?

Still loving this plugin! :+1:

SamPotts:
I think using `!default` would solve this? Sorry, I've always been a LESS person and have been meaning to move over to SASS but just haven't found the time. I'll check it out and get back to you.

Cheers :+1:

ivanvermeyen:
Thanks.

Based on this answer: http://stackoverflow.com/questions/12521197/sass-how-to-set-a-variable-to-equal-nothing the following seems to work for me:

```
// Contrast
@if lightness($plyr-controls-bg) >= 65% && not $plyr-control-color {
    $plyr-control-color: $plyr-gray-light;
}
@elseif not $plyr-control-color {
    $plyr-control-color: $plyr-gray-lighter;
}
@if lightness($plyr-control-bg-hover) >= 65% && not $plyr-control-color-hover {
    $plyr-control-color-hover: $plyr-gray;
}
@elseif not $plyr-control-color-hover {
    $plyr-control-color-hover: #fff;
}
```

PR #167, opened 2016-02-12
Allow player control colors to be overridden #158

ivanvermeyen:
The defaults for `$plyr-control-color` and `$plyr-control-color-hover` were overridden in the `if` structure, so an end user couldn't override the defaults themselves.

I deleted the initial defaults, and set the variables in the `if` structure as `!default` instead.

Hope this helps...


SamPotts:
Cheers :+1:

PR #168, opened 2016-02-17
Fixes for captions

gurupras:
Minor fixes for captions to better support manual rendering of captions.


SamPotts:
Thanks for this :+1:

PR #181, opened 2016-03-12
Dynamically set YouTube autoplay from config

shakeelmohamed:
The YouTube autoplay doesn't work without this change.

cc: @BlinkBP


SamPotts:
Thanks for this :+1:

Issue #201, opened 2016-05-06
CSS duplicate selectors


robinpoort:
I found two duplicate selectors in both the less and scss files:

The first one:
https://github.com/Selz/plyr/blob/master/src/scss/plyr.scss#L79-L81
and
https://github.com/Selz/plyr/blob/master/src/scss/plyr.scss#L110-L112

Second one:
https://github.com/Selz/plyr/blob/master/src/scss/plyr.scss#L449-L456
and
https://github.com/Selz/plyr/blob/master/src/scss/plyr.scss#L458-L464

Same goes for .less files. Want me to do a PR for this? Or is this something you rather do yourself real quick?

robinpoort:
I went ahead and created a PR :)

SamPotts:
Merged this 👍

PR #202, opened 2016-05-06
Removing duplicate selectors

robinpoort:
When using SCSS source files to compile, the duplicated selectors are not being merged. By removing the duplicated selectors in the source files the compiling will always result in non-duplicated selectors.

Issue: #201


SamPotts:
Cheers dude! 👍  I noticed this in the develop branch but cheers for sorting it.

robinpoort:
👍

PR #204, opened 2016-05-09
Broken progress bar styling IE10

robinpoort:
IE10 needs flex items to be at least `inline-block` elements. See: http://stackoverflow.com/questions/22008135/internet-explorer-10-does-not-apply-flexbox-on-inline-elements

I didn't compile the CSS file since I don't have Gulp installed and all. Added `inline-block` manually to the CSS file though. Might wanne re-compile but I don't think it's necessary :)


SamPotts:
Cheers 👍

robinpoort:
👍

Issue #206, opened 2016-05-10
Volume cursor invisible on IE11+


emmanuelclement:
This problem is already visible on https://plyr.io/ (volume cursor on video and audio)

emmanuelclement:
Seems like adding display: flex and flex: 1 to this definition fix the problem : 

@media(min-width:768px) {
  .plyr .plyr__volume{
    max-width:100px
  }
}

PR #207, opened 2016-05-10
Fix for issue #206

emmanuelclement:
Fix for Volume cursor invisible on IE11+

### DIFF
--- a/src/less/plyr.less
+++ b/src/less/plyr.less
@@ -636,7 +636,8 @@
         z-index: 2;
     }
     @media (min-width: @plyr-bp-screen-sm) {
-        display: block;
+        display: flex;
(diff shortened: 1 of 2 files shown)


src/less/plyr.less
```diff
@@ -636,7 +636,8 @@
         z-index: 2;
     }
     @media (min-width: @plyr-bp-screen-sm) {
-        display: block;
+        display: flex;
```
SamPotts: I'm pretty sure you don't need the `display: flex` but just the flex: 1 which could be specified on the parent selector

SamPotts:
I manually cherry picked this and merged it in. Cheers 👍

PR #209, opened 2016-05-11
Add CSS settings

robinpoort:
Adding three settings to be able to compile cleaner and faster CSS when compiling Plyr CSS yourself.
- Added setting to be able to disable `border-box` for everything if you already defined that in your own css
- Added setting to be able to disable `touch-action` for `a, button, input, label` if you already defined that in your own css
- Added setting to choose not to use `!important` for `plyr__sr-only` class. If you don't have any `span {}` or `button span {}` or `div span {}` etc. styling in your own css it's not a problem that the `plyr__sr-only` class doesn't have `!important`

All three settings are `true` by default so the CSS in the compiled CSS is exactly the same. The reason for the change in the CSS file is the fact that I grouped `clip` and `overflow` together at the top to be able to group the other values. See: https://github.com/Selz/plyr/commit/2a822d7b45e83972ee84d3b5118c027258d70f5e#diff-18cf65477dae5a253e0d3a854aeb66deL140


SamPotts:
Thanks for this 👍

robinpoort:
You're most welcome 👍 

Thanks for the awesome plugin! :D

Issue #210, opened 2016-05-12
Video: tooltips hidden on IE11


emmanuelclement:
On Internet Explorer 11, buttons are natively overflow:hidden (see user-agent CSS on http://www.iecss.com/ ).
(No overflow specified on IE Edge so you can see the tootips.)

Buttons are overflow:visible on others browsers.

note: normalize.css set overflow visible on button and input.

I'll make a pull request soon today to fix it.

SamPotts:
Merged this, cheers!

PR #211, opened 2016-05-12
Fix for issue #210

emmanuelclement:
adding overflow: visible in order to make button's tooltips visible on IE 11 (and normalize buttons behaviour).


src/scss/plyr.scss
```diff
@@ -651,7 +652,8 @@
         z-index: 2;
     }
     @media (min-width: $plyr-bp-screen-sm) {
-        display: block;
+        display: flex;
```
SamPotts: What's the reason for changing this to flex? Pretty sure `block` is fine?

src/less/plyr.less
```diff
@@ -652,7 +653,8 @@
         z-index: 2;
     }
     @media (min-width: @plyr-bp-screen-sm) {
-        display: block;
+        display: flex;
+        flex: 1;
```
SamPotts: As with below, don't think we need this to be `display: flex` and the `flex: 1` is already declared on the element outside of the media query

SamPotts:
Cheers for doing this, just left a couple of notes 👍

SamPotts:
I've manually merged this one (the button with overflow: visible) - I think the other issues were already resolved - cheers! 👍

PR #217, opened 2016-05-16
Added a hover effect for the large video play button

StelianAndrei:
I added a small hover effect to the button shown on the videos. It just swaps the blue and white colors around to give the user the sensation that something will happen when clicking that button. See the attached screenshot.

before-after


SamPotts:
Going to have to merge this one manually but will do. Cheers 👍

igoradamenko:
@SamPotts Can you wait with merge this, please? Today I will try find & suggest good hover effect for thumbs in #226 and I think it would be great to use similar effect. Color inversion does not work with thumbs. May be we'll be able to find shadow / border / opacity way to do similar hover effects.

I'll suggest something in 4-5 hours.

PR #226, opened 2016-05-18
Add hover effect for range thumbs

igoradamenko:
hover


SamPotts:
When I was designing the recent update to the UI, I did think about adding hover styles but concluded that it wasn't really necessary. If you think it is then maybe we can add something a little more subtle like an alpha transparent box-shadow thats nicely transitioned or perhaps scale the thumb a touch but I think flipping the color to the same as the progress bar is not something I'd go for...

igoradamenko:
@SamPotts No problem. I will play around with box-shadow / rgba and post here some versions. And we will be able to choose the best one :)

SamPotts:
Sounds good mate. You might want to update your branch as I changed `transition` to be `all .3s ease` so it won't auto merge. Sorry!

igoradamenko:
Hey! I've recorded four demos. Here they are:

First. Just scaling of thumb. Simple, clear, elegant:


Second. Thumb with slight shadow:


Third. 1st + 2nd:


Fourth. There are: box-shadow, scaling + border. Interesting effect, but..


But I think that 1st example is awesome. All plyr UI doesn't have shadows (right?) and they aren't needed here. What do you say?

robinkooli:
Yeah 1st example is awesome

SamPotts:
I'll add v1 in to the next release then - subtle is good 👍

robinkooli:
I meant this https://cloud.githubusercontent.com/assets/6537798/15412402/da319af6-1e2f-11e6-806d-c17a4cea2145.gif

SamPotts:
Yep. The first example as you said.

PR #237, opened 2016-05-22
Update readme.md

itaditya:
_thang_  was a less common word , keeping it simple by using 'thing' is I think good , also there were **SASS** support so why not add that too.


SamPotts:
SASS and LESS should be all caps.

SamPotts:
Cheers!

itaditya:
:)

PR #260, opened 2016-06-02
use $plyr-control-icon-size in sass file

cytec:
seems like the sass file was missed when $plyr-control-icon-size was first created :)


SamPotts:
Cheers 👍

PR #276, opened 2016-06-15
Add iPad to userAgent supported check

wilsonmsalberto:
When using an iPad, the "full" version wasn't working. 
By adding the "iPad" string to the user agent check it seems to solve the issue.

Cheers!

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -3279,7 +3279,9 @@
     function supported(type) {
         var browser = _browserSniff(),
             oldIE   = (browser.isIE && browser.version <= 9),
-            iOS     = /iPhone|iPad|iPod/i.test(navigator.userAgent),
+            iPhone  = /iPhone|iPod/i.test(navigator.userAgent),
+            iPad    = /iPad/i.test(navigator.userAgent),
+            iOS     = iPhone || iPad,
             audio   = !!document.createElement('audio').canPlayType,
             video   = !!document.createElement('video').canPlayType,
             basic, full;
@@ -3287,7 +3289,7 @@
         switch (type) {
             case 'video':
                 basic = video;
-                full  = (basic && (!oldIE && !iOS));
+                full  = (basic && (!oldIE && !iPhone));
                 break;
 
             case 'audio':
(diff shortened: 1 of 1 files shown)


SamPotts:
HTML5 Video is working fine as far as I'm aware. Users have reported issues with YouTube and Vimeo though. Are you seeing issues with HTML5 video?

src/js/plyr.js
```diff
@@ -3279,15 +3279,15 @@
...
-            iPhone  = /iPhone|iPod/i.test(navigator.userAgent),
+            iOS     = /iPhone|iPad|iPod/i.test(navigator.userAgent),
             audio   = !!document.createElement('audio').canPlayType,
             video   = !!document.createElement('video').canPlayType,
             basic, full;
 
         switch (type) {
             case 'video':
                 basic = video;
-                full  = (basic && (!oldIE && !iPhone));
```
SamPotts: I think this can just remain for iPhone only given videos can't be played inline on them *yet* (coming in iOS 10 by the sounds of it).

src/js/plyr.js
```diff
@@ -3279,15 +3279,15 @@
     function supported(type) {
         var browser = _browserSniff(),
             oldIE   = (browser.isIE && browser.version <= 9),
-            iPhone  = /iPhone|iPod/i.test(navigator.userAgent),
```
SamPotts: This might need to remain for the below.

wilsonmsalberto:
Yes, the issue has come up when playing video from Vimeo.

So, maybe we can check only for the iPad string, specifically on Vimeo

wilsonmsalberto:
I've updated the supported check to check for iOS user agents only on Vimeo.
I've also separated the iPad check, just in case it is needed afterwards.

I think this may avoid messing with the HTML5 Video that was working fine.

SamPotts:
I've merged this in manually since I had other changes. Thanks for helping out with it 👍

gino8080:
Can't see the merge, i have the same problem on ipad/iphone,
pointer-events 
 should I take another branch?
thank ou

SamPotts:
@gino8080 I'll have a look when I find some time

PR #287, opened 2016-06-25
Prefixed appearance:none was missing for Range Input

EnigmaSolved:
I'm not 100% certain if the `-moz` prefixed line is needed, but without the `-webkit` line I get funky results in Chrome.

### DIFF
--- a/src/scss/plyr.scss
+++ b/src/scss/plyr.scss
@@ -60,6 +60,8 @@
         padding: 0; 
         vertical-align: middle;
         
+        -webkit-appearance:none;
+        -moz-appearance:none;
         appearance: none;
         cursor: pointer;
         border: none;


SamPotts:
These prefixed declarations are handled using autoprefixer. I'd like to keep it that way as it manages which versions are required etc.

EnigmaSolved:
Ah, I understand. Yeah, I'm not currently using autoprefixer, but it's on my list (to do). Programming is not my day job (I'm actually a counselor), so I'm currently on an old Compass setup (which has ways of handling prefixing for me in SASS). But I'd like to migrate to using autoprefixer in the future.  I'm not sure what I'll do in the meantime though--I may just have to go through and switch the pertinent declarations to Compass versions so that I'll get prefixes where needed. 

It might be good to put a brief comment in the variables SASS file letting folks know that prefixing is expected to be handled by them using autoprefixer, as I only caught this because I was getting wonky results in Chrome.

SamPotts:
PostCss has a plugin for autoprefixer if there's any way you can use that? Sorry, I'd add the prefixed value to source but it means it's something else to maintain when the browsers eventually ditch the prefixes. 

I'll add a note in though, good suggestion! 👍

EnigmaSolved:
Yeah, no. I wouldn't want you to be adding them in manually--that'll drive you bonkers. :)

And thanks for the PostCSS mention--I've got to do some research on what route forward makes sense for me. I eventually want to update my whole build process, but you know how it is with time for such things.

SamPotts:
Totally. I'd love to devote more time to Plyr but struggle to find time. Cheers for the feedback 👍

EnigmaSolved:
You bet man--glad to help!

PR #288, opened 2016-06-25
Adding in missing prefixed appearance:none lines (for LESS)

EnigmaSolved:


SamPotts:
These prefixed declarations are handled using autoprefixer. I'd like to keep it that way as it manages which versions are required etc.

PR #326, opened 2016-08-04
Updated define to work with AMD imports

mkaziz:
Related to [issue 298](https://github.com/Selz/plyr/issues/298#issuecomment-236685838)


SamPotts:
Thanks 👍

Issue #318, opened 2016-07-22
Vimeo postmessage issue after destroying player


frumbert:
I have a vimeo player which I'm needing to destroy - but vimeo throws an error after the destroy.

`VM289 plyr.froogaloop.js:1 Uncaught TypeError: Cannot read property 'postMessage' of null`

Example:
https://jsfiddle.net/6okm62pf/1/

It consistently throws AFTER the destoy() since the vimeo video is still on the page.

If you remove the `.plyr` node itself right afterwards it seems ok. e.g.

    player_reference.destroy();
    var p = document.querySelector(".plyr");
    p.parentNode.removeChild(p);

but then the function isn't returning the node back to its original state (my case that's ok, but it's not what it says on the tin). :)

SamPotts:
Will take a look at this. I've just upgraded the Vimeo API to their new `player.js` which may help but I need to look at the `destroy()` method anyhow. Thanks for reporting 👍

frumbert:
On 1.8.12 I still have to destroy the actual plyr instance in order to make the video dissapear. If I then create a new instance, I get an uncaught typeerror - cannot set property 'embed' of undefined.

Example:
https://jsfiddle.net/6okm62pf/4/

1. Just run it and click destroy - the video temporarily goes away then comes back (Vimeo video seems to be calling back and recreating itself).
2. Click Destroy & RemoveChild - everything looks fine, no errors
3. Now click Create - you get the set property error, and plyr ins't properly set up.

I realise that if i'm changing source then I should be using `player.source` and giving it its new sources - but I can't make that work once/if the existing player is destroyed - https://jsfiddle.net/6okm62pf/7/ (clicking Create without first destroying the player still works)

SamPotts:
I'm looking at this now along with trying to clean up the object references a little more. Will be re-writing `.destroy()` as I think it's been forgotten a long the way and should be used by `.source()` anyway as a lot of the code is replicated.

Slgoetz:
Any solutions here?  I am having the exact same problem.  I am unloading the element on an Ajax page load then re-load in a video element only to get that same error. 
`cannot set property 'embed' of undefined.`

SamPotts:
Really sorry - I am working on it. Hoping I can get a fix out for this, plus a load of other issues this weekend. Have you tried the `.source()` method for updating instead of destroying and starting over?

SamPotts:
[...]

PR #329, opened 2016-08-05
Fix Vimeo's unload bug

jbourassa:
Hey,

This is in response to #318. The issue is that, for some reason, Vimeo's player still tries to send message using the iframe when removing the node. The trick is to unload it properly using `.unload()`, waiting until everything has been unloaded, and then remove the node. It is cumbersome, unfortunately.

-------------------

While digging I also tried to fix the destroy method to bring back the original DOM, but this requires more changes that I'm comfortable doing. Both YT and Vimeo are left in somewhat shady states after a `.destroy`. Here's my thought process / investigation:

- I'm assuming that `destroy` and `restore` should go in pair — a `restore` should bring back what has been `destroy`d.
- Plyr changes the actual DOM provided, so the most straightforward way to bring it back to its original state is cloning the node at the start.
- The passed in node gets modified _outside_ of Plyr's `init` method ([here](https://github.com/Selz/plyr/blob/master/src/js/plyr.js#L3415) and [here](https://github.com/Selz/plyr/blob/master/src/js/plyr.js#L3436)). This breaks my assumption that `destroy` and `restore` should go in pair.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -537,9 +537,13 @@ Here's a list of the methods supported:
     <td>Set the poster url. This is supported for the <code>video</code> element only.</td>
   </tr>
   <tr>
-    <td><code>destroy()</code></td>
-    <td>&mdash;</td>
-    <td>Destroys the plyr UI and any media event listeners, effectively restoring to the previous state before <code>setup()</code> was called.</td>
+    <td><code>destroy(...)</code></td>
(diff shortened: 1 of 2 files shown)


SamPotts:
Hey, thanks for this - I've manually merged this into some other refactoring work I've been doing. 👍

PR #330, opened 2016-08-05
Fix vimeo embed crashes post initial set

### LINKED ISSUE #318: Vimeo postmessage issue after destroying player

jbourassa:
It works on the first initialization and then crashes on later initializations because the first one ends up being async because [Vimeo's script is not loaded yet](https://github.com/Selz/plyr/blob/master/src/js/plyr.js#L1509-L1521)[1]. In a sync world, `plyr.container.ply` has not yet been set.

Since `plyr.container.plyr` should hold a reference to the `plyr` object, there is no point in referencing the container only to come back to the current instance. Thus, the order of execution doesn't matter anymore.

Fixes #318 

[1] I normally make all possilby-async-code always async, instead of sometimes async / sometimes sync, as it can lead to subtle bugs like this one

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -1556,9 +1556,6 @@
 
         // When embeds are ready
         function _embedReady() {
-            // Store reference to API
-            plyr.container.plyr.embed = plyr.embed;
-
             // Setup the UI if full support
             if (plyr.supported.full) {
                 _setupInterface();
@@ -2756,9 +2753,6 @@
             // Set aria title and iframe title
             config.title = source.title;
             _setTitle();
-
-            // Reset media objects
-            plyr.container.plyr.media = plyr.media;
         }
 
         // Update poster


src/js/plyr.js
```diff
@@ -1556,9 +1556,6 @@
 
         // When embeds are ready
         function _embedReady() {
-            // Store reference to API
-            plyr.container.plyr.embed = plyr.embed;
```
SamPotts: This line will remove access to the underlying YouTube or Vimeo API which some may find useful?

src/js/plyr.js
```diff
@@ -2756,9 +2753,6 @@
             // Set aria title and iframe title
             config.title = source.title;
             _setTitle();
-
-            // Reset media objects
-            plyr.container.plyr.media = plyr.media;
```
SamPotts: Again, this allows access to the media after a source change - otherwise the reference is to the old element

SamPotts:
I'm not sure what this actually fixes? I still get the post message bug when switching sources with these two lines removed. Thanks for looking into it though. It would not surprise me if the post message bug is a Vimeo thing. It seems calling `.unload()` makes no difference. Their Fragaloop (or whatever weird name they gave it) API was mega buggy.

jbourassa:
Hey Sam,

>Again, this allows access to the media after a source change - otherwise the reference is to the old element

Unless I'm missing something, I do not think this is correct. I understand that `plyr.container.plyr === plyr` (ie exact same reference), which means that assigning anything like so `plyr.container.plyr[whicheverProp] = value` is _exactly_ the same as `plyr[whicheverProp] = value`.

>It would not surprise me if the post message bug is a Vimeo thing. It seems calling .unload() makes no difference.

The postmessage is indeed inside Vimeo, it happens if you remove the iframe before vimeo is done cleaning up. I think they probably should handle it differently, but #329 allows not having this message by letting developers remove the DOM node after vimeo's `unload` is complete.

SamPotts:
Well, we had an issue without the media being *reset* on the container although the solution could have been wrong. 

I'm looking at the `.setup()` / `.destroy()` methods now as I've never been 100% happy with them (and destroy is broken currently) - I'll take another look and go from there. I'll also include your other changes.

jbourassa:
Thank you @SamPotts!

SamPotts:
Hey, thanks for this - as per the other PR I've manually merged this into some other refactoring work I've been doing. 👍

PR #333, opened 2016-08-10
Use `forEach` instead of `for (var x in foo)`.

abuiles:
Given the supported browsers, it is safe to use `forEach` which is a
better version of `for (var x in foo)`.

Since `for` iterates on the `enumerable properties`, it can lead to
unexpected behaviors when you are extending the array prototype. As an
example, check the following ember-twiddle, which show all the other
returned properties when using `for x in foo` in Ember.

https://ember-twiddle.com/bac28d42c800c4dffc4253f66afd0cac?openFiles=controllers.application.js%2C.

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -25,7 +25,7 @@
     'use strict';
 
     // Globals
-    var fullscreen, 
+    var fullscreen,
     scroll = { x: 0, y: 0 },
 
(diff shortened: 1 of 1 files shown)


SamPotts:
Thanks for this - I was doing a load of refactoring work so I've just manually merged this as part of that. 👍

PR #341, opened 2016-08-20
Support deezer player

tamdao:


SamPotts:
Thanks for spending time on this. I think I'm going to create some form of plugin system to allow for extra providers like this to be plugged in. I'll come back to this shortly. 👍

tamdao:
@SamPotts Thanks. Let me know if the feature is ready

PR #360, opened 2016-09-02
Add fullscreen container

amowu:
Why add this feature? because I want to active fullscreen with plyr's parent element!

For example:

```
<div class=".video-container">
	<!-- some other component for video -->
	<!-- original plyr container -->
</div>
```

Now I can use:

```
plyr.setup(element, {
	selectors: {
		fullscreen: {
			container: '.video-container'
		}
	}
});
```

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -65,6 +65,9 @@
                 container:      null,
                 wrapper:        '.plyr__controls'
             },
+            fullscreen: {
+                container:      null,
+            },
             labels:             '[data-plyr]',
             buttons: {
                 seek:           '[data-plyr="seek"]',
@@ -122,8 +125,7 @@
         fullscreen: {
             enabled:            true,
             fallback:           true,
-            allowAudio:         false,
-            container:          null
+            allowAudio:         false
         },
         storage: {
             enabled:            true,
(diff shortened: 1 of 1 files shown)


src/js/plyr.js
```diff
@@ -122,7 +122,8 @@
         fullscreen: {
             enabled:            true,
             fallback:           true,
-            allowAudio:         false
+            allowAudio:         false,
+            container:          null
```
SamPotts: Is this needed? I think you meant to put this in selectors?

Parasoul: I agree. Put it in the selectors as fullscreenContainer. Then use config.selectors.fullscreenContainer at line 904.
Fullscreen isn't in selector, so you can't access it from selectors if i'm right.

src/js/plyr.js
```diff
@@ -65,6 +65,9 @@
                 container:      null,
                 wrapper:        '.plyr__controls'
             },
+            fullscreen: {
+                container:      null
+            },
```
Parasoul: Was especting fullscreenContainer : null but it's better this way.

src/js/plyr.js
```diff
@@ -915,7 +917,7 @@
                     _log((nativeSupport ? 'Native' : 'Fallback') + ' fullscreen enabled');
 
                     // Add styling hook
-                    _toggleClass(plyr.container, config.classes.fullscreen.enabled, true);
+                    _toggleClass(plyr.fullscreenContainer, config.classes.fullscreen.enabled, true);
```
Parasoul: You have to use plyr.selectors.fullscreen.container according to line 68.
Same for line 2113/2124/2740.

Parasoul:
Re-open it it was good, i misstaken myself.

PR #364, opened 2016-09-07
now hides cursor just like somethingtube

richard512:


sampotts:
Just going to leave this for now. Cheers 👍

PR #376, opened 2016-09-23
Added thumbnails on seek tooltip

tdaines:
I added the option to show thumbnails when the user hovers over the seek progress bar.  A new `showThumbnails` option was added along with a `thumbnail` event in which the thumbnail image source and dimensions may be set.

My only concern is when the thumbnail is far to the left, it goes past the edge of the video like so:
capture

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -85,7 +85,7 @@
             var instances = plyr.setup({
                 debug:              true,
                 title:              'Video demo',
-                showThumbnails:     false,
+                showThumbnails:     true,
                 tooltips: {
                     controls:       true
                 },
--- a/demo/index.html
+++ b/demo/index.html
@@ -73,16 +73,69 @@
         </main>
 
         <!-- Plyr core script -->
-        <script src="../dist/plyr.js"></script>
-
-        <!-- Docs script -->
-        <script src="dist/demo.js"></script>
+        <script src="../src/js/plyr.js"></script>
 
         <!-- Rangetouch to fix <input type="range"> on touch devices (see https://rangetouch.com) -->
         <script src="https://cdn.rangetouch.com/0.0.9/rangetouch.js" async></script>
 
         <!-- Sharing libary (https://shr.one) -->
         <script src="https://cdn.shr.one/0.1.9/shr.js"></script>
-        <script>if(window.shr) { window.shr.setup({ count: { classname: 'btn__count' } }); }</script>
+        <script>if(window.shr) { window.shr.setup({ count: { classname: 'btn__count' } }); }
+
+            var instances = plyr.setup({
+                debug:              true,
+                title:              'Video demo',
+                showThumbnails:     false,
+                tooltips: {
+                    controls:       true
+                },
+                captions: {
+                    defaultActive:  true
+                },
+            });
+
+            // Plyr returns an array regardless
+            var player = instances[0];
+            player.on('ready', function(event) {
+                console.log('ready');
+            });
+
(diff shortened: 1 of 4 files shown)


SamPotts:
Thanks - I'll add this into the next release

SamPotts:
There was a suggestion that we do this through a `<track>` element with `type="metadata"` which could be used for this?

tdaines:
Can you link me to this suggestion?  My understanding is that text tracks are for timed data, triggered by the position of the player, whereas thumbnails are triggered by the user and the `onmouseover` event.

SamPotts:
There were a couple of other ideas in there:
https://github.com/Selz/plyr/issues/46

The `<track>` could be another option - seems cleaner to pull the URLs from a file. Perhaps we just use events initially and expand it later.

tdaines:
It's your call.  I don't have a file of thumbnail URLs for my assets though, so pulling URLs from a file doesn't help me much.  Maybe I'm in the minority on that though.  I'm happy to help implement whatever.

gpapin:
Hello I was looking for a new web player just for this particular feature ! If I can add my pinch of salt,
I would say that both ideas could be improved. Getting the thumbnail from a video or a sigle file seems very inefficient. I think a better approach would be to load a sprite with all you thumbnails in if and move the sprite accordingly.

I think that jw player found a good way of solving this specific issue : https://support.jwplayer.com/customer/en/portal/articles/1407439-adding-preview-thumbnails

I would very much like this to be implemented so let me know if I can help :)

Mte90:
any hope for that feature?

jqEmprendedorVE:
This feature is implemented?

gehaktmolen:
I tried this out, but i can't see in your code where you actually display the thumbnail. Maybe forgot to commit it? Anyway i made a quick thing to at least show the thumbnail. Could probably use some debounce method and variable dimensions. in _updateSeekTooltip:
`
var canvas = document.createElement('canvas');
canvas.width = 160;
canvas.height = 90;
canvas.getContext('2d').drawImage(plyr.media, 0, 0, 160, 90);
plyr.progress.seek.thumbnail.src = canvas.toDataURL();
`
Also maybe better to put the flag for enabling this within the seeker object? Instead of enableThumbnails?

tdaines:
@gehaktmolen I figured getting thumbnails can be fairly specific to each individual, so I left out the actual displaying of them from the PR.  You can see my test code for displaying the thumbnails [here](https://github.com/sampotts/plyr/pull/376/commits/a7ffaae4ada7dcfeea42ae0f966b81a6a07fbd35#diff-3e22c50371c76d64ea9738dbc46a6e57R135).  I just set the `src` and dimensions of the image and let the browser download the image.

@gpapin While I agree an image sprite would be ideal, I don't have access to any videos with thumbnail sprites nor does my company create them for our videos.  If demand is there, I can work on a new PR to support thumbnail sprites once the new version of plyr comes out.

sourovroy:
Please merge this pull request, This is a very useful feature.

sampotts:
This PR was created against v2 so it'd need to be re-written for v3 before it can be merged.

tdaines:
If I get some free time soon, I'll work on a v3 version of this PR.  Maybe as a plugin or something.

friday:
@sourovroy posted this on our Slack:
> i have already written for testing purpose and one of my project here
https://github.com/sourovroy/plyr-test

I can't say for sure since it's not a clear case, but I do think this would be better as a Plyr plugin, which is something we don't yet support. I've been looking at it, but there are architectural challenges.

ufukomer:
So, what is the right way to help this issue?

sampotts:
Closing this one as we have the functionality now. Cheers for the hard work on this thread/PR 👍

PR #385, opened 2016-09-29
Added internal support for multiple tracks

dacostafilipe:
Here's internal support for multiple tracks. With the extended API `setCaptionIndex` you can switch between the available tracks.


SamPotts:
This is perfect timing! I'm just working on the UI for this. I'll release this, this weekend along with playback speed and quality (for YouTube / Vimeo) 👍

SamPotts:
Hey mate, I'm having to merge this one manually but I have a question. Is there any reason we don't use the `srclang` attribute instead of relying on the index in the source? `_setCaptionIndex` could then be `_setCaptionLang` or whatever and just pass the matching language? Happy to do the change (since I'm in there) if there wasn't a compelling reason to use index instead?

dacostafilipe:
Yeah, I had the same idea but what if you have two tracks with the same language? For example an english translation and an english translation with sound effects for deaf people? 

Or what if people do not pay attention to the `srclang` attribute?

SamPotts:
That's true, it's just relying on the source code order seems fragile to me. I'll have a think about it. Maybe it's fine as is. 

Sam

dacostafilipe:
I've looked this up and I still think that using the index is the best way to do it internally.

But you are right that the setCaptionIndex may not be clear enough for everybody.

How about adding a "getCaptionTracks" that would help the users choose the right track?

The output could be something like this :

[
	{
		track 	: <Element>
		index 	: 0,
		kind 	: 'captions',
		src		: '...',
		lang 	: "en",
		label 	: "English"
	}
] 

Also, how about adding an event for caption selection/change?

SamPotts:
That sounds ideal - that way you can choose the one that suits, yeah. If you want to make that change, can you do it in the develop branch as that's what I'm doing the UI for this in (not working yet).

dacostafilipe:
Kinda messed up with the branches on the fork, so I'll need to clean them up first. I will then properly create a new branch based on develop to work on this.

dacostafilipe:
I've pushed the changes to my develop branch : 

https://github.com/dacostafilipe/plyr/commit/c1b25c0e97d47f32b0d4a3fd7f2521dbbf1a71d1

PS: I'm still having sync issues with my fork, sry for that :/

SamPotts:
OK mate, thanks for that. I'll manually merge your changes over don't worry. Sorry yeah, git can be fiddly at the best of times with forks and PR's. Don't worry too much 👍 Thanks again.

SamPotts:
Closing this one for now

PR #421, opened 2016-11-10
Replace player.js with custom-build version, and move script tag to last

JodiWarren:
https://github.com/Selz/plyr/issues/420

1. Add a custom-built version of Vimeo's player.js. The only change is to change the rollup config's format to `umd` from `iife`

2. Insert the `<script>` tag as the last `<script>`, rather than the first. I did this as part of my bugfixing, but it seemed like a good option anyway.

### Task list

- [ ] Gulp build completed


src/js/plyr.js
```diff
@@ -295,8 +295,9 @@
 
         var tag = document.createElement('script');
         tag.src = source;
-        var firstScriptTag = document.getElementsByTagName('script')[0];
-        firstScriptTag.parentNode.insertBefore(tag, firstScriptTag);
+        var scripts = document.getElementsByTagName('script');
+        var lastScriptTag = document.getElementsByTagName('script')[scripts.length - 1];
```
SamPotts: Can't this just be:
```
var lastScriptTag = scripts[scripts.length - 1];
```
Save a lookup?

SamPotts:
Can this not be fixed in the Vimeo lib itself? I'd rather not modify their code and then host that modified file in your repo.

src/js/plyr.js
JodiWarren: Absolutely should be, good catch.

JodiWarren:
I'm not sure how this can be fixed in Vimeo's lib. My suggestion would be to take the rebuilt player.js and host it via the CND that you used to use for Froogaloop.

You can build your own version from https://github.com/vimeo/player.js/

SamPotts:
When you say CND do you mean CDN?

OK maybe we can build it into the gulp build somehow. I'll take a look.

peirix:
Had a look at this yet? It's needed to have vimeo playback work when loading plyr with requirejs, as described in this GitHub issue from vimeos player.js: https://github.com/vimeo/player.js/issues/79

friday:
I'm going to close this since it's very outdated and can't be merged, wasn't a sustainable solution to begin with. The link was to a private repo which didn't update and which we didn't have control over. Also: vimeo now supports amd (in case they didn't before).

Example of how to import vimeo in require.js:
http://jsbin.com/filogakaqu/edit?js,console

If someone needs to make alterations to the code to support this use case (I think it is still the case, but I don't use require.js, so I wouldn't know), please create a new pull request to the `develop` branch. 👍 

And thanks for the effort!

PR #425, opened 2016-11-17
fixes bug where cleanUp would be called twice

sebastiancarlsson:

https://github.com/Selz/plyr/issues/416

When destroying a Vimeo video a timeout is also set that calls cleanUp() after 200ms, apparently because "Vimeo does not always return". I assigned an ID to this timeout so that it can be cleared when cleanUp() is executed to prevent cleanUp() from executing twice.

### Task list

- [ ] Gulp build completed

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -25,7 +25,7 @@
     'use strict';
 
     // Globals
-    var fullscreen, 
+    var fullscreen,
     scroll = { x: 0, y: 0 },
 
     // Default config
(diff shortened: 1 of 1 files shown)


sebastiancarlsson:
Ok so in Sublime Text 2 I have the "trim_trailing_white_space_on_save" set to true (and you should too) so there are a lot of changes that are just removal of whitespace.

The lines with actual changes to the code are 3260 and 3276.

SamPotts:
Thanks for this. I'll enable that setting. Cheers 👍

sebastiancarlsson:
Awesome

PR #427, opened 2016-11-24
Fix cannot scroll after destroyed

amowu:

In single page application, switch router when plyr is fullscreen mode, page cannot scroll, because `document.body.style.overflow` is `hidden`

### Task list

- [ ] Gulp build completed


SamPotts:
Cheers 👍

PR #458, opened 2016-12-16
fix: parse webVTT time NaN

amowu:

If VTT time format is `00:00:15.792`, it will be work,
but some VTT time has `00:15.792` format, it can not work.

### Task list

- [ ] Gulp build completed


SamPotts:
Cheers 👍

PR #509, opened 2017-02-14
Fix on non clickable parent due to icon

pchrysa:

When I set a custom iconUrl then the "Settings" button becomes non clickable if clicked on top of the icon. This does not happen when the default iconUrl (https://cdn.plyr.io/2.0.10/plyr.svg) is set. I had to set a "pointer-events: none;" on less and sass in order for the event to properly work.

### Task list

- [x] Tested on [supported browsers](https://github.com/Selz/plyr#browser-support)
- [x] Gulp build completed


SamPotts:
Manually merged to `develop`

Issue #115, opened 2015-08-24
Loop Functionality for API - Loop video n times


MaryCoding:
Overall a great tool. However I dont see a way to set a loop counter for video through API. I know there is an html5 video attribute `loop`. This will make the video play infinitely. However, is there a way to set limits to loop. For example, set video to only play 5, 10 or 15 times. An icon they could click to set that number. Any ideas? Looking forward to your thoughts about this.

SamPotts:
I'll see if I can work on something. I need to do something for playback speed as well so this could be done at the same time.

MaryCoding:
:+1:  Awesome.

PR #510, opened 2017-02-14
Added key bindings for loop

pchrysa:

#115 

Added key binding events for loop. These keybindings can be changed through config as well. As default keys I have set:

l - Toggle between Loop All/No Loop event
i - Sets the start marker of the loop
o - Sets the end marker of the loop

### Task list

- [x] Gulp build completed 
- [x] Tested on Chrome and Firefox


SamPotts:
Cheers for this - I'm going to make a change to hardcode the keycodes for now as the others are hardcoded so we're consistent

PR #545, opened 2017-03-25
Add blank.mp4 to dist files

silverwind:

https://github.com/Selz/plyr/pull/504

For a complete "offline" experience, it's necessary to self-host the file, so include it in the dist files for that purpose.


silverwind:
A cleaner approach could to include the file in the form of a data-uri, in which case, the `blankUrl` option can be dropped, and `plyr.js` would gain 2395 bytes. Shall I go for it?

SamPotts:
I had that before - there was a reason I reverted to the remote file. I can't remember why...

silverwind:
Okay, then let's keep it like this. I'm fine with it.

SamPotts:
Merged to `develop` so will go with `v3.0.0` when it's done 👍

PR #563, opened 2017-04-19
removed vimeo id parsing because it is not necessary

stormrockwell:
This parser causes issues with different vimeo url formats such as: 
https://vimeo.com/channels/staffpicks/194361952

I propose removing the id parser for Vimeo urls completely instead of fixing it. Vimeo's JS API accepts either an id or URL.


stormrockwell:
If you prefer I can do a better regex to pull the id but I don't think that is needed

SamPotts:
This looks good - cheers 👍

SamPotts:
Merged into `develop` so it'll go with `v3.0.0`

SamPotts:
Ah looks like I'll have to revert this. It doesn't work passing a URL unfortunately. Would have been good to test before hand. I get this:

image

SamPotts:
All good - it looks like the source video was removed. Whoops.

stormrockwell:
Haha alright. Got scared I broke it for a second. Thanks for the merge

PR #577, opened 2017-05-01
fix for svg4everybody conflicting with plyr

brandon-pereira:
https://github.com/Selz/plyr/issues/407

Currently if you have svg4everybody on your page and you also use Plyr the icons won't appear. This is because Plyr is hardcoded to always pull SVGs via absolute urls on IE... Since svg4everbody is making them accessible from the local url we need to have an option to enable this.

### Task list

- [x] Tested on [supported browsers](https://github.com/Selz/plyr#browser-support)
- [x] Gulp build completed


brandon-pereira:
Good idea, updated to reflect. Note: If you're using webpack (like I am) you'll need to manually assign svg4everybody to the window.

sampotts:
Cheers for this 👍  Have retrofitted to v3 too

PR #617, opened 2017-07-12
[bugfix] target is null exception

oemueller:

when using 'play-large' without a 'play' button, the statement target = target[0]; caused an error. So it is necessary to first check the value of target before accessing [0]

### Task list

- [x] Gulp build completed


sampotts:
Cheers for this 👍

Issue #587, opened 2017-05-23
For the video players some of the player area is not clickable just above the player controls


dineshb-qa:
- [ ] Issue observed on https://plyr.io

### Expected behavior 
For the video players,  all the player area should be clickable so that user can play/pause the video consistently across the screen

### Actual behavior
For the video players some of the player area is not clickable just above the player controls

### Environment

- Browser:
- Version:  Chrome 58, Firefox 53
- Operating System: Win 8

Players affected:
- [ ] HTML5 Video
- [ ] YouTube
- [ ] Vimeo

### Steps to reproduce 
- Hit the URL : https://plyr.io
- Observe that : On start screen or when the video is playing the player area  just above the player controls is not clickable

jjui:
Yup, got this issue too. 
I'm hiding the controls, but the area behind is not clickable for start/stop.
This may be resolved with just a higher `z-index` on `.plyr__controls` and full height of the toggle. 
Will see what I can find...

jjui:
Update
```
.plyr--video .plyr__controls {
  pointer-events: none;
  & > * {
  pointer-events: all;
  }
}
```
Will make a untested PR as I cannot think of any issue with the merge :)

sampotts:
This is fixed in a recent PR and latest version.

YLaido:
I remember that it used to be available to pause the video by clicking on the area above the controls, which is unfeasible now. I've tried on the demo site with Firefox and Chrome but no luck. Do you have any solution to this? 
Thank you.

PR #654, opened 2017-09-09
Enable start/stop toggle on whole element

jjui:
https://github.com/sampotts/plyr/issues/587

Disable pointer-events on `.plyr__controls` and add the events to all children so the whole element can be clicked. This will also solve the none-clickable hidden control area when controls is hidden.

### Task list
Not tested, no build, sorz
- [ ] Gulp build completed


sampotts:
Cheers. This is actually resolved in v3 as the controls slide out of the element.

PR #667, opened 2017-09-26
Fix 'TypeError: Cannot read property 'kind' of null' in captions feat…

friday:
`player.captions.currentTrack` is null in Chrome 61 when there are no captions, so the caption `supported` detection triggers an error (the detection still actually works though).

screen shot 2017-09-27 at 00 33 09

I didn't create any issue since it's in the "There will be bugs" a.k.a. develop branch

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -2448,7 +2448,7 @@
                 });
 
                 // Check if suported kind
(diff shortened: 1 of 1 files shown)


sampotts:
Cheers 👍

PR #670, opened 2017-09-27
Avoid errors when initiating plyr without settings (v3)

friday:
Plyr 3 (develop branch) won't work if you initiate without the "settings"-control:

Example:
```js
new Plyr(video, {controls: ['play', 'progress']})
```

Will result in multiple errors trying to access properties of `undefined`.

This PR checks that `setCaptionsMenu()` and `setSpeedMenu()` aren't run unless settings are enabled and those sections are enabled.
It also checks that `form` exists before trying to access properties and ensures `updateSetting` won't break if the pane doesn't exist.

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -1948,7 +1948,9 @@
             player.elements.controls = controls;
 
             //setLoopMenu();
-            setSpeedMenu();
+            if (utils.inArray(player.config.controls, 'settings') && utils.inArray(player.config.settings, 'speed')) {
+                setSpeedMenu();
+            }
 
             return controls;
         }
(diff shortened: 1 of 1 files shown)


sampotts:
Thanks for the help 👍 I'm trying to get to it myself but struggling for time as per usual.

friday:
No worries. These were small and easy fixes, and I'm glad to help anyway :)

I think v3 is working really well for my purposes so it will end up in production soon (I really only need the ability to switch captions for now). I'm about to switch from html5 video to using HLS.js though, so maybe I'll revise that statement ;)

I'm a little bit confused about the css. While I can "fix" things easily, maybe I better let you handle this since it's "work in progress" and I can't know your intentions (see font, settings-icon and options line-height in picture).

(Could create an issue for this too if you want)

screen shot 2017-09-27 at 16 17 04

Issue #331, opened 2016-08-05
Uncaught (in promise) DOMException: The play() request was interrupted by a call to pause()


joseluisq:
When the video is stopped and I'm dragging the `plyr__progress` bar to the next minute, 
it appears this error:

`Plyr v1.8.12`

![image](https://cloud.githubusercontent.com/assets/1700322/17452652/e1bedc22-5b36-11e6-8698-d2cc5e5e65c7.png)

SamPotts:
This seems to be a Chrome issue by the looks of things:
https://bugs.chromium.org/p/chromium/issues/detail?id=593273

filoghost:
There is a workaround, not pretty but it should work: http://stackoverflow.com/questions/36803176/how-to-prevent-the-play-request-was-interrupted-by-a-call-to-pause-error

SamPotts:
Closing this and waiting on a Chrome fix

neutrosolutions:
From html video TAG remove autoplay
and use this code to play.
`hls.attachMedia(video);
                hls.on(Hls.Events.MANIFEST_PARSED, function () {
                    video.play();
                });
hls.on(Hls.Events.ERROR, function (event, data) {
                    if (data.fatal) {
                        switch (data.type) {
                            case Hls.ErrorTypes.NETWORK_ERROR:
                                // try to recover network error
                                console.log("fatal network error encountered, try to recover");
                                hls.startLoad();
                                break;
                            case Hls.ErrorTypes.MEDIA_ERROR:
                                console.log("fatal media error encountered, try to recover");
                                hls.recoverMediaError();
                                break;
                            default:
                                // cannot recover
                                hls.destroy();
                                break;
                        }
                    }
                });`

alejandroiglesias:
I would like to reopen this because the aforementioned Chrome issue says the error is fixed for when calling `HTMLMediaElement.pause` before `HTMLMediaElement.play`, but calling `HTMLMediaElement.play` and then `HTMLMediaElement.pause` while the media is loading still throws the error.

```
// fixed
media.pause();
media.play();

// won't fix; still shows the issue.
media.play();
media.pause();
```

`HTMLMediaElement.play` is async and returns a promise. The error complains about having called `HTMLMediaElement.pause` before the playback started.

I would like to quote to the Chromium issue owner:
[...]

PR #676, opened 2017-10-03
Native play can throw promise exception

pionl:
Discussed in #331 issue. This will make exception to use console.log

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -2075,7 +2075,7 @@
         // Issue #331 - chrome raises exception if pause is called before media is playing
         function catchChromeException (promise) {
             if (promise) {
-                promise.catch((error) => {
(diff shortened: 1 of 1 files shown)


friday:
The Chrome-bug was fixed. This is happening due to race-conditions and/or invalid states. See: https://stackoverflow.com/questions/36803176/how-to-prevent-the-play-request-was-interrupted-by-a-call-to-pause-error for how to really avoid the rejection. You don't need to handle `The play() request was interrupted by a call to pause()` though. You can just ignore the promise. Uncaught promise rejections are not errors/exceptions, though I agree they can be annoying.

A better solution imo would be to return the original play() event (the promise) in plyr's play method so users who would want to handle it could.

sampotts:
This should be sorted in v3 👍

PR #678, opened 2017-10-06
Add URL as a possible input option to Vimeo player

yuriykuzin:
This is reasonable to process correctly private Vimeo links, like https://vimeo.com/25345658/fe46e209ac - that contains not only vimeo id, but also private key.

So, from this point, we can pass a vimeo url to `data-video-id` attribute, like we do in our project:

```
<div #plyrPlayer
        attr.data-type="{{ externalType }}"
        attr.data-video-id="{{ currentUrl }}">
</div>
```

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -654,10 +654,10 @@
         return url.match(regex) ? RegExp.$2 : url;
     }
 
-    // Parse Vimeo ID from url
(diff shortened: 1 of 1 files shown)


sampotts:
Cheers. I need to merge this to `develop` somehow for the v3 release. Will take a look 👍

PR #693, opened 2017-10-25
v3: Add instanceof wrapper to avoid TypeErrors

friday:
`x instanceof y` will result in an error if `y` is undefined. For example `new VTTCue(0, 2, 'Hello') instanceof window.VTTCue` will throw something like "TypeError: Right-hand side of 'instanceof' is not an object" in browsers that don't support the VTTCue API.

This PR adds a wrapper that checks that x and y are "truthy" first, and return false if not, (or if x isn't an instance of y).

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -373,10 +373,10 @@
                 return input !== null && input instanceof Event;
             },
             cue: function(input) {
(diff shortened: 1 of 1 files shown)


sampotts:
Cheers dude! 👍

PR #694, opened 2017-10-25
Restore volume stepping via key up/down

friday:
Restores v2's behavior with 10 steps when changing volume with up/down keys (currently in develop  it jumps directly to max/min).


sampotts:
You're on fire mate :1st_place_medal:

friday:
:)

PR #696, opened 2017-10-26
Enable speed controls using Vimeo API

yuriykuzin:

If we enable speed controls in vimeo options, we can then adjust playback rate: `player.getEmbed().setPlaybackRate(speed)`


sampotts:
This is coming in v3 (which is unfinished and currently in `develop`) so I'll close this for now but thanks for your time on it 👍

yuriykuzin:
Hi @sampotts! The reason why I requested this PR is to enable working with speed through native youtube/vimeo api, which I already use in production

sampotts:
Sure - will add it in then 👍

PR #700, opened 2017-10-27
fix #684 memory leaks issues after destroy.

### LINKED ISSUE #684: destroy method not working as expected
### Expected behaviour 
Would expect plyr instances to be destroyed after calling `instances.forEach(function(instance) { instance.destroy(); }`

### Actual behaviour
Unforunately the js instances are still there, the method only destroys the DOM elements. Because of the instance still existing, the entire plyr DOM elements can still be found as detached elements in the memory.

### Environment
- Browser: Chrome
- Version: latest  
- Operating System: Windows
- Version: latest 

Players affected:
- [x] HTML5 Video
- [x] HTML5 Audio
- [x] YouTube
- [x] Vimeo

### Steps to reproduce 
Instantiate player(s), call `destroy()` method on it and then try and access the instance. For example call the `play()` method on it. Even though the initial DOM elements are destroyed, you're still able to play the audio file...

### Relevant links
https://github.com/sampotts/plyr/issues/77

ashkanhosseini:
https://github.com/sampotts/plyr/issues/684

Cleaning up plyr.container when calling destroy method in order to let GC remove all event listeners and prevent memory leak. 

- Tested on Chrome 61.0.3163.100
- Gulp build completed

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -519,11 +519,11 @@
     }
 
     // Unbind event
-    /*function _off(element, events, callback, useCapture) {
+    function _off(element, events, callback, useCapture) {
         if (element) {
             _toggleListener(element, events, callback, false, useCapture);
         }
-    }*/
+    }
 
     // Trigger event
     function _event(element, type, bubbles, properties) {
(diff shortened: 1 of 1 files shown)


sampotts:
Cheers 👍

PR #703, opened 2017-11-01
Allow Plyr.setup event listeners to be set up as separate event listeners

gurupras:
This allows the setup listeners to do things like
preventDefault()/stopImmediatePropagation() and have them work

There is an [issue comment](https://github.com/sampotts/plyr/pull/141#issuecomment-341026257)

Earlier, if a `listener` was specified during `Plyr.setup()`, we set up one event handler with both the provided listener (`userListener`) as well as the `defaultListener`.

Now, each of these is attached as a separate event listener (`userListener` being attached first).
This ensures that `preventDefault()`/`stopImmediatePropagation()` are handled correctly.

### Task list

- [ ] Gulp build completed

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -472,13 +472,21 @@
 
     // Bind along with custom handler
     function _proxyListener(element, eventName, userListener, defaultListener, useCapture) {
+        if(userListener) {
+            // Register this before defaultListener
+            _on(
+                element,
+                eventName,
+                function(event) {
+                    userListener.apply(element, [event]);
+                },
+                useCapture
+            );
+        }
         _on(
             element,
             eventName,
             function(event) {
-                if (userListener) {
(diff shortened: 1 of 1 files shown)


sampotts:
Cheers - will retrofit this to `develop`

sampotts:
Ported over to `develop` like this

```javascript
// Trigger custom and default handlers
const handlerProxy = (event, customHandler, defaultHandler) => {
    // Execute custom handler
    if (utils.is.function(customHandler)) {
        customHandler.call(this, event);
    }

    // Only call default handler if not prevented in custom handler
    if (!event.defaultPrevented && utils.is.function(defaultHandler)) {
        defaultHandler.call(this, event);
    }
};

// Play
utils.proxy(this.elements.buttons.play, 'click', this.config.listeners.play, togglePlay);
```

Issue #697, opened 2017-10-27
Vimeo controls displaying behing Plyr controls


lukeyouell:
Issue is visible on plyr.io website (https://plyr.io/#vimeo)

### Expected behaviour 

Plyr controls should only be visible

### Actual behaviour

Both Vimeo and Plyr controls are visible

### Environment

- Browser: Chrome (latest)
- Version:  Plyr 2.0.16

Players affected:
- Vimeo

### Steps to reproduce 
- Embed Vimeo as per docs

### Relevant links

https://plyr.io/#vimeo

aymenpeak:
it happened to me too !

weskhaled:
me too

sampotts:
Argh, Vimeo really is crap. I think it might be time to drop support. There's no way to hide their controls via the API so we had a kinda hack to hide them off screen (iframe was 200% height and then centered in the viewport) but clearly that's no longer working. Either they found it on someone's site or they discovered Plyr. A really downer. 

I think the options are either we drop support or keep it but using their controls which would offer an inconsistent experience and frankly not as good looking. 

It's a shame as they have some good quality videos but when it comes to embedding that content on your own website, YouTube far exceeds them and even they have their downsides.

yuriykuzin:
As a temporary solution, I'm placing non-transparent plyr control over vimeo's one. And only in case of non-fullscreen, because in fullscreen the is no such problem for now.

Also, had to add defaultDelay to my fork of plyr to show plyr controls longer than vimeo's one.

![screen shot 2017-10-27 at 2 26 39 pm](https://user-images.githubusercontent.com/1729556/32102003-2954243c-bb23-11e7-8582-07933cab43b6.png)

BrunoDeBarros:
This has happened to me as well, but it is possible to turn the controls off in the embed settings: 

https://i.28hours.org/20171031-173541-1509.png

It's not perfect, and they then add a "play" button to the centre of the video, but it works well enough to make continuing to use Plyr viable. And the good news is Plyr can get rid of that Vimeo hack to hide controls now, since it's no longer of any use.

sampotts:
Is it possible to turn them off for everyone? I was planning to just use their "native" controls for now and add an option to the config to turn on the custom controls.

BrunoDeBarros:
[...]

PR #704, opened 2017-11-03
Fix issue #697

jonnitto:
Unfortanly `transparent` is not a supported option in the [vimeo player](https://github.com/vimeo/player.js#embed-options). So I add `&transparent=0` after the iframe is initialized

- [x] Gulp build completed


sampotts:
Closing this as it's fixed via the `<iframe>` solution. Thanks for the work though 👍

PR #705, opened 2017-11-04
Add .editorconfig file

friday:
Add .editorconfig

Why: Some editors will fail to identify tab width otherwise (like Sublime Text 3 in my case).


sampotts:
Cheers mate! 👍

PR #763, opened 2018-01-17
Beta with ads - Reloading ads on new video

gehaktmolen:
Made sure its possible to reload the adsmanager with new ads when the video is done or when ads are done. This will make it easier to implement ads within a playlist sort of feature. Moved the ads container outside of the video wrapper so its on the same level as the controls container, making it easier to move the ads in front of controls or behind controls. Removed the logic for starting the ad by tapping the ads container. From what i can see is that you already made it possible to start the ad from plyr itself?

### Task list

- [x] Gulp build completed

### DIFF
--- a/src/js/plugins/ads.js
+++ b/src/js/plugins/ads.js
@@ -33,7 +33,6 @@ class Ads {
      * Get the ads instance ready.
      */
     ready() {
-        this.time = Date.now();
         this.adsContainer = null;
         this.adDisplayContainer = null;
         this.adsManager = null;
@@ -366,11 +365,15 @@ class Ads {
     play() {
         const { container } = this.player.elements;
 
-        // Initialize the container. Must be done via a user action on mobile devices.
-        this.adDisplayContainer.initialize();
+        if (!this.adsManagerPromise) {
+            return;
+        }
 
         // Play the requested advertisement whenever the adsManager is ready.
         this.adsManagerPromise.then(() => {
+            // Initialize the container. Must be done via a user action on mobile devices.
+            this.adDisplayContainer.initialize();
+
             try {
                 if (!this.initialized) {
 
--- a/src/js/plugins/ads.js
+++ b/src/js/plugins/ads.js
@@ -51,20 +51,20 @@ class Ads {
         this.events = {};
         this.safetyTimer = null;
 
+        // Set listeners on the Plyr instance.
+        this.setupListeners();
+
         // Setup a simple promise to resolve if the IMA loader is ready.
-        this.adsLoaderResolve = () => {};
         this.adsLoaderPromise = new Promise((resolve) => {
-            this.adsLoaderResolve = resolve;
+            this.on('ADS_LOADER_LOADED', () => resolve());
         });
         this.adsLoaderPromise.then(() => {
             this.player.debug.log(`[${(Date.now() - this.time) / 1000}s][IMA SDK] adsLoader resolved!`, this.adsLoader);
         });
 
         // Setup a promise to resolve if the IMA manager is ready.
-        this.adsManagerResolve = () => {};
         this.adsManagerPromise = new Promise((resolve) => {
-            // Resolve our promise.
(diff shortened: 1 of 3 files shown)


gehaktmolen:
I did some removals or "big" changes. I tried to keep them within separate commits, so its easier to see. Hope you like the changes.

gehaktmolen:
Added mid-roll cue markings within the progress bar: http://prntscr.com/i1pvks  similar to the youtube video player. Hope you like it.

sampotts:
Hey @gehaktmolen , thanks for your help with this. Can you make sure you've setup prettier, eslint and stylelint so your code follows the standards set out in the config files. e.g. `if(` should be `if (` (missing space).

src/js/plugins/ads.js
```diff
@@ -378,17 +428,47 @@ class Ads {
...
+                this.onAdError(adError);
             }
         });
     }
 
+    /**
+     * Resume our video.
+     */
+    contentResume() {
+        this.player.debug.log(`[${(Date.now() - this.time) / 1000}s][IMA SDK]`, 'Resume video.');
```
sampotts: You can enable timestamps in your browser. Also, you can remove the fullstops from console messages to follow the standard. So this would become just `Resume video`.

src/js/plugins/ads.js
```diff
@@ -378,17 +428,47 @@ class Ads {
...
+        // Play our video.
+        if (this.player.currentTime < this.player.duration) {
+            this.player.play();
+        }
+    }
+
+    /**
+     * Pause our video.
+     */
+    contentPause() {
```
sampotts: Can't this just be `pause` ?

src/js/plugins/ads.js
```diff
@@ -378,17 +428,47 @@ class Ads {
...
+                // VAST response.
+                this.onAdError(adError);
             }
         });
     }
 
+    /**
+     * Resume our video.
+     */
+    contentResume() {
```
sampotts: Can this just be `resume` ?

gehaktmolen: Yeah no problem. I just figured it would be confusing if someone were to call: `player.ads.resume();` instead of `player.ads.contentResume(); as you don't resume the ad, but the actual video.

sampotts: Good point - maybe make it `resumeContent()` then?

gehaktmolen: Done

src/js/plugins/ads.js
gehaktmolen: I think i understand what you mean. Hope my recent change has resolved this.

src/js/plugins/ads.js
gehaktmolen: Done

gehaktmolen:
Hope the re-formatting is what you expect it to be. I'm having trouble setting up the correct linting within phpstorm. Within my own projects I use google formatting, as its default within my IDE (lazy :P). It doesn't have an airbnb config unfortunately so I had to set some settings manually.

I usually have eslint as part of my Grunt build tasks, so when i make a mistake it simply won't build the project. I assumed this was the case with Plyr as well, as when i do a pull request there is a task (checkbox) within the description text-area to successfully run the "build". My train of thought was: successful build === good formatting :P

src/js/plugins/ads.js
```diff
@@ -21,14 +29,16 @@ class Ads {
...
+     * Get the ads instance ready.
+     */
     ready() {
         this.time = Date.now();
         this.adsContainer = null;
         this.adDisplayContainer = null;
         this.adsManager = null;
         this.adsLoader = null;
         this.adsCuePoints = null;
-        this.currentAd = null;
```
gehaktmolen: Removed `this.currentAd`. Noticed it was not being used. That ok>?

sampotts:
Thanks for your help! I should make gulp do the formatting somehow, you're right. 👍

Issue #775, opened 2018-01-30
Plyr is not defined  (v3.0.0-beta.10)


ispal:
- [x] Issue does not already exist
- [ ] Issue observed on https://plyr.io

### Expected behaviour 
When importing library from NPM using Webpack and `import Plyr from "plyr"`, I expect everything to work normally using examples on the site. It works if I include the script using script tag

### Actual behaviour
I get error in the console saying `ReferenceError: Plyr is not defined`.

Error comes from this code on dist version according Dev Tools

```
dispatchEvent: function(e, t, i, n) {
   if (e && t) {
       var s = new CustomEvent(t,{
                       bubbles: !!c.is.boolean(i) && i,
                       detail: Object.assign({}, n, {
                          plyr: this instanceof Plyr ? this : null
                       })
                    });
        e.dispatchEvent(s)
       }
 },

```
### Environment

- Browser: Chrome
- Version:  63.0.3239.132
- Operating System: MacOS 
- Version: High Sierra

Players affected:
- [x] HTML5 Video
- [ ] HTML5 Audio
- [x] YouTube
- [ ] Vimeo

### Steps to reproduce 

```
import Plyr from 'plyr';

const player = new Plyr('#player', {
    autoplay: true
});
player.source = {
    type: 'video',
    sources: [
      {
        src: 'bTqVqk7FSmY',
        provider: 'youtube'
      }
    ]
};
```

sampotts:
This should be sorted in 3.0.0-beta.11

Maqsyo:
Thanks for this fast fix!

PR #776, opened 2018-01-30
Import plyr.js into utils.js file. This should fix error #775 when tr…

ispal:
…ying to import using npm

https://github.com/sampotts/plyr/issues/775

Import Plyr class directly to utils.js file. I tested this with using the beta version 10 on my local setup and directly modifying utils.js file and added `import Plyr from "./plyr"`. I also builded `dist/plyr.js` file in a forked repository and copy pasted the result to my app and it worked.

### Task list

- [ ] Gulp build completed

### DIFF
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -2,7 +2,6 @@
 // Plyr utils
 // ==========================================================================
 
-import Plyr from './plyr';
 import support from './support';
 import { providers } from './types';
 
@@ -588,7 +587,7 @@ const utils = {
         const event = new CustomEvent(type, {
             bubbles: utils.is.boolean(bubbles) ? bubbles : false,
             detail: Object.assign({}, detail, {
-                plyr: this instanceof Plyr ? this : null,
+                plyr: (window.Plyr && this instanceof window.Plyr) ? this : null,
             }),
         });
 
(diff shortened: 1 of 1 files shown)


sampotts:
Wouldn't it be better to do something like this?

```javascript
this instanceof window.Plyr ? this : null
```

ispal:
Then I will get the following error `Uncaught TypeError: Right-hand side of 'instanceof' is not an object`.

Then it should be
`window.Plyr && this instanceof window.Plyr`

Checking if the Plyr exists in the window before checking the instance.

sampotts:
Ah, just realised I should have just used the helper at the top of the file `utils.is.plyr(this) ? this : null`

ispal:
Would love to see fix for this in the next beta because it's kind of blocking me currently.

sampotts:
Can you try v3.0.0-beta.11 ? I just pushed with a change to use `utils.is.plyr()` instead as it follows the standard to use the `is` helper.

ispal:
I tried v3.0.0-beta.11 and that fixed it. Thanks! 👍

sampotts:
Awesome. Thanks for your help with it 👍

Issue #771, opened 2018-01-29
[v3] Captions not showing in IE


friday:
Captions are not showing in IE11 / Win 10. Possibly also Edge browser (not tested)
* No related error is shown.
* The `.plyr_captions` element still exists, and it has a span child which is empty. If you manipulate the span and add text it will appear correctly styled as in other browsers.

- [x] Issue does not already exist
- [x] Issue observed on https://plyr.io/beta

### Environment

- Browser: Internet Explorer
- Version:  11
- Operating System: Windows
- Version: 10

Players affected:
- [x] HTML5 Video
- [ ] HTML5 Audio
- [ ] YouTube
- [ ] Vimeo

friday:
(will try to fix this myself, but not right now)

friday:
Additional info: `document.querySelector('video').textTracks[0].cues.length` should be 7 for https://plyr.io/beta. It is in other browsers, but it's 0 in IE11. So this is the root cause it seems. Possibly format/mime-type related.

sampotts:
It's really weird. IE11 *should* support VTT files and I swear it did in testing before. I'll keep digging. Thanks for taking a look 👍

friday:
Yes, VTT does work in IE11. I've also managed to get it to work now. Not sure exactly how. I'm pretty sure the only thing I did in effect was upgrade to the latest beta version, but it still doesn't work on https://plyr.io/beta. Nothing about the demo code looks wrong however.

I'll keep the issue open until we understand what's causing it.

friday:
If it helps, this page (uses beta 12 with #780 applied) works with IE11: https://workflow.studiotogo.com/view/59db9312345a8234c27c809b (beware the bad speech recognition)

sampotts:
Done some more digging and it appears IE11 doesn't seem to like loading captions from a remote URL for whatever reason. Using a relative URL like you've done seems to work fine for me too. I'm thinking of moving v3 to master and using GitHub Pages to host plyr.io. Maybe that'll help solve it.

sampotts:
Looks like IE11 never worked with CORS VTT:

- https://github.com/videojs/video.js/issues/2073
- https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8832169/

I'm not going to restore all that code to load the captions with AJAX and the extra logic so might just add a note in the docs about it.

friday:
Nice detective work! Could never have guessed this was the cause.

[...]

PR #780, opened 2018-02-06
Fix harmless but annoying IE error 'IndexSizeError'

friday:
Note: This is not a fix for #771. It's just avoiding an error thrown frequently by IE11 trying to access `activeCues[0]` when `activeCues` is empty.

This doesn't happen in other browsers or in IE11 with normal arrays afaik. Really weird.


sampotts:
Cheers mate! 👍

PR #799, opened 2018-02-27
Fixed setQualityMenu bug for youtube videos

sndrgb:
Hi, 
I fix a simple bug when you play/pause youtube video.
It always retry to setQualityMenu after state changing, but for example, I omit settings control from my video options.

What do you think?

### Task list

- [x] Gulp build completed


friday:
I think it looks good, but removing the gulp build would make the PR and git history a lot less noisy and less likely to conflict.

My opinions are not official though.

sampotts:
I'll just retrofit this one as it's easier

sampotts:
I've moved the check to the `setXMenu` methods themselves incase they get called from elsewhere. Will push shortly.

Issue #537, opened 2017-03-17
Vimeo Seek not working


frumbert:
- [ ] Seems similar to https://github.com/Selz/plyr/issues/208 (still open)

### Expected behaviour 
1. Calling `seek` on the `ready` event should set the start position to that point then start playback

### Actual behaviour
Vimeo always starts at the beginning.

### Environment

- Browser: Chrome 
- Version:  56 / 57
- Operating System: MacOS Sierra
- Version: 10.12.3

Players affected:
- [X] Vimeo

### Steps to reproduce 
- include cdn versions of plyr code.
- put ONE div on the page, pointing to a vimeo id
- var _player =plyr.setup()[0];
- _player.on("ready", function (event) { event.detail.plyr.seek(60); });

### Relevant links
Internally plyr calls the setCurrentTime() method on the vimeo player.

Vimeo documentation: https://github.com/vimeo/player.js#setcurrenttimeseconds-number-promisenumber-rangeerrorerror

Pen: https://codepen.io/frumbert/pen/Rpjypx

weskhaled:
Try:      
  plyerplyr[0].on('playing', function (event) {
          event.detail.plyr.seek(50);
  });

sampotts:
I can't seem to replicate this on v3 so assuming it's fixed? Feel free to re-open if that's not the case 👍

Issue #208, opened 2016-05-11
Forcing the player to start at a specific position?


mattmurtaugh:
I'm working on a project that keeps track of how far along a user is on the video, and pushing that information to the database every 5 seconds. 

I'm unable to get the seek method to work without input from the user. I am able to have a link that calls the seek method, and that works fine, but when I try to call the seek method when the file loads, nothing happens. 

This works: 

`$('.seek').click(function(e) {
	e.preventDefault();
	var starttime = parseInt($('.plyr').attr('data-start'));
	player.seek(starttime);
});`

But when I try to call `player.seek(starttime);` by itself nothing happens. 

Am I overlooking a trick to doing this? Or is there an easier way to specify where the video should start playing?

Thanks

mattmurtaugh:
After playing around with this for a little while I found that if I set a very short delay, I don't have any issues with this at all. It's working like I want it too, but is there a better way to accomplish this than waiting 300 milliseconds before triggering the seek method?

SamPotts:
You might be able to listen to the 'ready' event and then make the call to `seek`? 
https://github.com/Selz/plyr#events

epnlespieux:
@mattmurtaugh i am very interested by this feature : i would like to split a long interview into small pieces with a direct acccess to the particular part as we can do with youtube. Can you show some code  for me to begin please ?
(sorry for my poor english, french am I)

mattmurtaugh:
I tried to listen for the 'ready' event, but it doesn't seem to be work how I thought it would when I first tried it. When I'm done with the project, I might revisit this idea.. But for now here is my solution:

`setTimeout(function(){
	player.seek(starttime);
},300); `

Replace 'starttime' with a variable that contains your start time or the time itself. You may also need to adjust the 300 milliseconds to something a little higher depending on how long it takes Plyr to load on your server. I found that I could use 250, but I bumped it up a little bit to make sure it loaded.

I'm far from the best programmer, so there is probably another way to accomplish this that is more efficient, but it seems to work for me.

skitterm:
I was looking for the same functionality. I put the seek call inside the ready event. YouTube works (it pops me to that spot in the video), but Vimeo doesn't (it starts from the beginning. I stuck a setTimeout of 1s and then called seek, and it worked for Vimeo as well... but not for html5 video. Also, I had to click "play" on Vimeo videos a couple of times before I could see both the video playing and the time progress going.

nOurs88:
Hi all,

[...]

PR #805, opened 2018-03-13
Fix player seeking not working for Vimeo

nikola3244:
#208 #537

First off, I am not much experienced with JavaScript development, so this solution might or might not be correct. 

I've just moved the Plyr load function to execute when Vimeo player has loaded (and thus has duration value). Seeking on `ready` event did not work because duration value was `0`, and based on my assumption - it's because Vimeo player didn't fetch the data yet. 

So, this code change fixes the issue with seeking.

Tested on Chromium on Arch Linux.

### Task list

- [ ] Gulp build completed

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -1930,9 +1930,6 @@
             plyr.media.paused = true;
             plyr.media.currentTime = 0;
 
-            // Update UI
-            _embedReady();
-
             plyr.embed.getCurrentTime().then(function(value) {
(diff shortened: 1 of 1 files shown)


sampotts:
Can you verify this is still an issue with v3 and if so then raise another PR for that? Cheers 👍

nikola3244:
@sampotts Sure. v3 is `beta` or `develop`?

tocsinDE:
v3 is in master by now.

PR #820, opened 2018-03-20
Fix fast-forward control

saadshahd:
fast-forward control doesn't work.

`controls.createButton` expects `forward` key not `fast-forward` https://github.com/sampotts/plyr/blob/bb7eea27e513bf33dd4a4d846b3234b37a5be48d/src/js/defaults.js#L286

### Task list

- [x] Gulp build completed


sampotts:
I think `fast-forward` makes more sense doesn't it?

saadshahd:
@sampotts Yes I believe so too, for sake of consistency. I'll do it.

saadshahd:
@sampotts I tried to unify its name to be `fastForward` but doing so will introduce breaking changes to the module exported API.

sampotts:
I just meant the controls side of things.

saadshahd:
I updated the controls side only @sampotts

saadshahd:
and you could check the whole API changes from here https://github.com/sampotts/plyr/compare/master...saadshahd:fix/unify-fast-forward-naming?expand=1

sampotts:
Cheers 👍

PR #863, opened 2018-04-04
Fix loading data-plyr-config when initiating Plyr without any options

friday:
`utils.extend()` works by iterating itself over each argument until the argument is no longer an object, in which case the iteration is ended. So `utils.extend({}, defaults, undefined, moreOptions)` will ignore `moreOptions`.

While this might be where the real issue is (depending on the intention), I didn't want to assume this was unintentional.


sampotts:
Awesome cheers 👍

Issue #865, opened 2018-04-04
[v3] Settings menu depends on speed being enabled


friday:
If I initiate Plyr without speed in settings, the setting icon doesn't show:

```js
new Plyr(video, {settings: ['captions']});
```

This line in `setSpeedMenu` is what enables the icon, but it's only being executed from `setSpeedMenu`, requiring speed to be enabled:
https://github.com/sampotts/plyr/blob/ee79c46145578c77fbf37c93389a3c7e7a950b0c/src/js/controls.js#L728-L729

friday:
#866 fixes this for html5, but not vimeo. Youtube presently only supports `speed` I believe, so it can't be tested.

sampotts:
This should be resolved in `v3.2.0` but please let me know if not 👍

friday:
Great! It does work as expected for my use case now anyway :)

PR #866, opened 2018-04-04
Remove settings dependency on 'speed'

friday:
### Link to related issue
#865

Moved the call to `checkMenu` from `setSpeedMenu` to the main Plyr constructor. 
In my tests the settings icon will show with either captions or speed in it, but not if it's an empty array. I haven't tested further than that.

I'm not fully sure this is safe, and I guess you might have a different solution in mind. So please be critical.

### DIFF
--- a/src/js/controls.js
+++ b/src/js/controls.js
@@ -725,9 +725,6 @@ const controls = {
         const toggle = !utils.is.empty(this.options.speed) && this.options.speed.length > 1;
         controls.toggleTab.call(this, type, toggle);
 
-        // Check if we need to toggle the parent
-        controls.checkMenu.call(this);
-
         // If we're hiding, nothing more to do
         if (!toggle) {
             return;
--- a/src/js/controls.js
+++ b/src/js/controls.js
@@ -530,6 +530,7 @@ const controls = {
     // Update the selected setting
     updateSetting(setting, container, input) {
         const pane = this.elements.settings.panes[setting];
+        const tab = this.elements.settings.tabs[setting];
         let value = null;
         let list = container;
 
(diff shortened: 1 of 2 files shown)


friday:
No, you're right. It doesn't!

sampotts:
It's so complex these days 😢

friday:
I managed to get it working for vimeo, but not youtube (as you foresaw). I'm also not sure the initial change to put it in the main constructor was  good idea as opposed to putting it somewhere in the html5 helper, vimeo and youtube respectively.

In addition to this, the demo doesn't like it when I toggle from one the other options to html5 video (settings icon disappears), which indicates it's not the right place.

Don't be afraid to close the issue, since I may not be on the right track. Perhaps checkMenu should be split up in `initiate` (before all data is available) and `verify` (once it is)?

friday:
Moved it to ui.build and fixed another symptom (typeerror) of the same unsupported use case. It now works perfectly for html5 video/audio, but the behavior is unchanged for vimeo and unknown for youtube (unless I'm mistaken, youtube settings only support speed currently so it can't be tested).

To replicate the vimeo issue, `settings` must contain `captions`, but not `speed`. Then the settings icon doesn't show up (like before).

For vimeo, this is the same as the current behavior (without the error), so it's not a regression. I haven't had the time to dig deeper so I think it's better to merge this (if you approve) and leave the issue open.

sampotts:
I think I have this sorted - just had to add it to the other menu methods

friday:
Or inside `toggleTab` perhaps. I thought about that but forgot about it.

sampotts:
This should be resolved in `v3.2.0` but please let me know if not 👍

PR #878, opened 2018-04-09
Allow the quality to be a string

philipgiuliani:
In my case I needed the qualities to be represented as string, because one of them is `audiodescription`.

I removed the `Number()` calls and replaced them with a new function called `mapQualityUnit`. It converts numbers to numbers, and keeps strings to be strings.

### Task list
- [x] Gulp build completed

### DIFF
--- a/src/js/html5.js
+++ b/src/js/html5.js
@@ -5,6 +5,14 @@
 import support from './support';
 import utils from './utils';
 
+function mapQualityUnit(input) {
+    const resolution = parseInt(input, 10);
+    if (utils.is.number(resolution)) {
+        return resolution;
+    }
+    return input;
+}
+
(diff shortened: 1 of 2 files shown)


friday:
To me this seems like a complicated way to end up not knowing what the type is anyway. Would rather have it switching the type consistently to string, or not enforcing a type. Then instead adapting `getBadge()` and any other places relying on quality being a number.

PS: I'm not the maintainer, so you may want to wait for Sam's opinions

friday:
@jacoz126 Learn how to solve your own problems [instead of being our problem](http://slash7.com/2006/12/22/vampires/).

philipgiuliani:
I close this in favour of a different approach i'm working on.

PR #887, opened 2018-04-11
Add i18n support for "Normal" value in speed options

danielsarin:

In speed options, if value is 1 then "Normal" is shown instead. This pr adds a new "normal" key to i18n options so that it's possible to translate that string.

### Task list

- [ ] Gulp build completed


sampotts:
Thanks mate 👍

PR #888, opened 2018-04-11
Safer check for active caption

Antonio-Laguna:
Some browsers don't seem to support `className` as a string that can be used with `indexOf` so they throw when using captions like this:

`Uncaught TypeError: a[b].target.className.indexOf is not a function`

I've just swapped that by `classList.contains` which should be safer and also (IMHO) clearer.

### Task list

- [ ] Gulp build completed


sampotts:
Thanks! 👍

Issue #893, opened 2018-04-12
[IE10] Unable to get property 'split' of undefined or null reference


nicolasthy:
### Expected behaviour
Have Plyr to work on IE10

### Actual behaviour
It throws an error in the console
`Unable to get property 'split' of undefined or null reference`

### Environment

- Browser: Internet Explorer
- Version: 10
- Operating System: Windows
- Version: 10

### Steps to reproduce
- Run any website with plyr on IE10

### Solution
The error comes from the `defaults.js` file where `window.navigator.language` is undefined.
I found a solution that should do the work:
```javascript
...
// Captions settings
captions: {
    active: false,
    language: window.navigator.language && window.navigator.language.split('-')[0],
},
...
```

friday:
Looks solid. I don't have any IE10 testing setup any longer, but [MDN confirms it](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language#Browser_compatibility). Maybe you could make this a PR?

nicolasthy:
I don't have access to contributions, had the fix ready but couldn't push on a new branch 😅

friday:
Only maintainers can create new branches. To create a PR as a non-maintainer you fork the repo, push your changes to a new branch and press "New pull request" from your branches list next to the branch you want to PR.

You can also just browse to the correct file and make the changes in the github text editor, which will fork Plyr and create a PR for you, but this way is probably best to only use for documentation changes since you can't test before.

nicolasthy:
Oh thank you ! I'm used to contribute to repositories in which I have permission to push on new branches.

nicolasthy:
I did some more testing, seems like `captions.language` is needed elsewhere in the code. Could we give it a default value when `window.navigator.language` is null?
Something like this:
```javascript
// Captions settings
captions: {
    active: false,
    language: window.navigator.language ? window.navigator.language.split('-')[0] : 'en'
},
```

sampotts:
I've merged the PR but will fallback to `navigator.userLanguage` for IE... 

```javascript
language: (navigator.language || navigator.userLanguage).split('-')[0]
```

sampotts:
This should be resolved in `v3.2.0` . Thanks for your help 👍

PR #895, opened 2018-04-12
Fix IE10 split error

nicolasthy:
https://github.com/sampotts/plyr/issues/893

On IE10, Plyr throws the error `Unable to get property 'split' of undefined or null reference`. This fixes the case when `window.navigator.language` is null and can't use the `split()` function.

### Task list

- [ ] Gulp build completed

### DIFF
--- a/src/js/defaults.js
+++ b/src/js/defaults.js
@@ -115,7 +115,7 @@ const defaults = {
     // Captions settings
     captions: {
         active: false,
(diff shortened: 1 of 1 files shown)


sampotts:
I think rather than default to english we can get `navigator.userLanguage` as that's the IE equivalent. I'll merge and update though.

PR #903, opened 2018-04-17
Show captions even if toggle button is omitted from controls

### LINKED ISSUE #901: Captions won't show if "captions" is omitted from controls
If controls are defined in settings, excluding "captions", then captions won't show at all, and switching between languages in the settings menu has no effect since `.plyr__captions` is hidden.

You can override this easily by forcing `.plyr__captions` to `display: block`.

Example:
https://jsfiddle.net/dv3j41go/9/

### Environment
- Browser: Chrome 65, Firefox (tested with these, but I don't think it matters)
- Operating System: osx/el capitan

friday:
Fixes #901

Just removed the conditions to check the button. Switching via settings or `player.toggleCaptions()` now works without the toggle button, as well as getting the right settings from either localStorage or the settings passed to Plyr.

This makes a lot more sense to me and I don't think the old check should be replaced with something else like checking if either the toggler or caption setting (with languages) elements exists. Better to respect user settings.

### DIFF
--- a/src/js/captions.js
+++ b/src/js/captions.js
@@ -250,11 +250,6 @@ const captions = {
 
     // Display captions container and button (for initialization)
     show() {
-        // If there's no caption toggle, bail
-        if (!utils.is.element(this.elements.buttons.captions)) {
-            return;
-        }
-
         // Try to load the value from storage
         let active = this.storage.get('captions');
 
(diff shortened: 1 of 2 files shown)


sampotts:
Cheers for this and the other PR!

Issue #946, opened 2018-05-08
Playback not working on mobile (click obscured by poster elem)


mhluska:
### Expected behaviour

Expect video to play when clicking play on iOS mobile Chrome.

### Actual behaviour

Video does not play. Play button animation happens.

### Environment

- Browser: Chrome
- Version: 66.0.3359.122
- Operating System: iOS
- Version: 11.3

### Steps to reproduce

- Click play on player

sampotts:
Which video type? I can't replicate on https://plyr.io for HTML5, YouTube or Vimeo.

mhluska:
The video type is webm. The first `<source>` tag is webm and the second is mp4 as a fallback.

sampotts:
As I say I can't replicate on the demo site. Could you link to the player with the issue? Any errors in the console perhaps?

mhluska:
I'm seeing this error in sentry: `The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.`

friday:
@mhluska Your fix will be included in the next plyr release. In the meanwhile you have to keep that in your css 👍

sampotts:
@mhluska @friday It's now released in v3.3.7 👍

medoingthings:
It seems like this fix was removed again. Came across the same issue with plyr on iPhone where the (progressive iframe) YouTube player would not be playable, because the poster layer „eats” the tap. Will use the workaround, but am curious why `pointer-events: none;` is gone.

plyr-version: 3.6.2

PR #950, opened 2018-05-08
Poster fixes

friday:
A couple of css fixes to the poster

1. Letterbox the poster rather than stretching it. This is how html5 video posters normally behave.
2. Add `pointer-events: none;` as suggested by @mhluska. I haven't reproduced #946 but think this should be safe either way.

### DIFF
--- a/src/sass/components/poster.scss
+++ b/src/sass/components/poster.scss
@@ -6,7 +6,7 @@
     background-color: #000;
     background-position: 50% 50%;
     background-repeat: no-repeat;
-    background-size: 100% 100%;
+    background-size: contain;
     height: 100%;
     left: 0;
     opacity: 0;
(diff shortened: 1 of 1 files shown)


sampotts:
Using `contain` won’t stretch a smaller image of course which I think the default does (I need to check again). It’s probably just `100%` (stretched horizontally) but this didn’t match the default behavior when I tested.

friday:
> Using contain won’t stretch a smaller image 

It does actually. You can try it yourself.

From the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/background-size):
> If the background-size is contain or cover:
> While preserving its intrinsic proportions, the image is rendered at the largest size contained within, or covering, the background positioning area. If the image has no intrinsic proportions, then it's rendered at the size of the background positioning area.

friday:
If you by "stretch" mean scale up that is. Otherwise I'm not sure how to interpret the comment.

sampotts:
Sorry, you are correct 👍

Issue #954, opened 2018-05-09
YouTube seeking with with seek tooltip and custom controls


jamesrockett:
### Expected behaviour
When I hover over the timeline on a YouTube video, I see the time code pop up in a tooltip. If I click on that spot in the timeline, the video should seek to the time code being displayed in the tooltip.  

### Actual behaviour
The time code displayed in the seek tooltip is not the same one that the video actually seeks to. It is inconsistent and off by approximately 15-20 seconds every time.

### Environment
- Browser: Google Chrome  
- Version: 66.0.3359.139
- Operating System: Windows 10 Pro
- Version: 1709

### Steps to reproduce
- Create a default, new Plyr instance with **custom controls** (HTML from the example in controls.md) using this video: **https://www.youtube.com/watch?v=FbfffBa9cFw** (any video should show it but a longer one makes it more obvious)
- Hover over the video's timeline and you should see a 'seek' tooltip showing the time code of that position on the timeline
- If you click, the video will go to a different position than what was shown in the tooltip
- If you click and drag, you will see that the tooltip time code does not match the current time

Hopefully this can be reproduced, I've noticed it on multiple machines. Thank you for your hard work, this is a great library!

friday:
I'm not getting this with the plyr demo (https://plyr.io/#youtube) with Chrome 66 on a Mac. Are you getting the issue on the demo?

friday:
Correction: It's happening on the demo as well.

friday:
I think it's a rounding error. The diff gets bigger the further to the right you click, and it's happening for html5 and vimeo too.

jamesrockett:
That makes sense, I first started noticing it on really long videos (1hr+)

sampotts:
The cause of this is the `<input type="range">` and my calculation of the position for the tooltip. The issue is that the handle never overlaps the end of the track so you need to know the width of the handle itself and account for a varying amount of that as you get closer to the extremities. Unfortunately you can't determine the width of the handle with JavaScript as it's in the shadow DOM. I wonder if it's possible to do by creating an offscreen `<input type="range">` that has `width: 0` then calculate the rendered width. I'll take another look. Someone else raised a similar bug.

friday:
Fixed in v3.3.8

Issue #808, opened 2018-03-14
Tooltip time has broken positioning in IE11


denilsonc:
Tooltip time not working in ie11/window 7

Português Brasil:
A Dica da ferramenta não funciona usando iframe no ie11, fiz testes no chrome, opera, firefox e não tive nenhum problema.
 Ao mover o mouse sobre a barra de progresso o pre view time fica totalmente a direta da barra mas o tempo continua normal.

Até então não sei de onde vem o problema, acho que o problema esta vindo de pageXOffset mas não tenho como verificar.

friday:
Please keep it in English (you seem to know enough English to use online translation and verify later).

denilsonc:
I think it's a bug!

when I enter full screen the browser read the progress bar with 6px wide.
http://www.christophercurrie.com/technology/2014/03/20/internet-explorer-11-fullscreen-bug.html

sampotts:
Translated:

> The tooltip does not work using iframe in ie11, I did tests in chrome, opera, firefox and had no problem.
> By moving the mouse over the progress bar the pre-view time is totally in the direct bar but the time remains normal.
> Until then I do not know where the problem comes from, I think the problem is coming from pageXOffset but I can not verify.

Will take a look.

friday:
I can't reproduce this.

denilsonc:
with full screen?

friday:
It works with or without fullscreen in my setup.

friday:
Closing, since this cannot be reproduced with the given instructions. Both the tooltip time (since #958) and position works in IE11, with or without fullscreen.

Issue #687, opened 2017-10-13
Seek tooltip is not always an accurate indicator of seek on click


Moiahbabi:
- [yes] Issue does not already exist
- [yes] Issue observed on https://plyr.io

### Expected behaviour 
while video paused, and seeking with mouse click , tooltip should show exact time as the video current time display.

### Actual behaviour
based on video duration, tooltip time and video current time are different by few second, in plyr.io 1s. see [img 01].

It seems to be due to the track thumb css properties. so I tried to give it a width of zero, with no borders
and it seems to give a correct time display for both tooltip and video current time. see [img 02].

### Environment

- Browser: (Chrome, Firefox, IE, Opera).
- Version:  
- Operating System: Windows 10
- Version: 1703

Players affected:
- [yes ] HTML5 Video
- [yes] HTML5 Audio
- [yes ] YouTube
- [yes] Vimeo

### Steps to reproduce 
- Load video
- pause video
- while paused, seek video by mouse click to a point further in the video.
- check tooltip time and current time. 

### Relevant links
[img 01]
![ezgif-5-b686d3fa15](https://user-images.githubusercontent.com/26636470/31553722-72a2668a-b04c-11e7-827f-d1f370a8b9cf.gif)

[img 02]
![2017-10-13 1](https://user-images.githubusercontent.com/26636470/31552667-35b26f02-b049-11e7-8317-0b635705b0aa.jpg)

sampotts:
This is a known issue and is super tricky - in the middle of the scrubber it's pretty accurate but as you go left or right of center the accuracy drops. It's worst at each extreme. The problem is the way the tooltip value is calculated is based on the mouse offset from left to right but the range doesn't quite work like that as the *edge* of the handle for the range input is only inline with the *edge* of the track rather than the center of the handle being in the *centre* of the *end* of the track.  I'll look into it but as I say, it's a tricky one to do.

friday:
Fixed in 3.3.8

PR #958, opened 2018-05-11
Fix the seek tooltip time difference from seek time

friday:
Hacky proposed fix for #687 #954 (and perhaps #808, but I don't understand it) using the sass var `$plyr-range-thumb-height` and negative margins to make the container the same size as the area the center point of the range thumb can move.

Tested in Chrome, Firefox and Safari.
Firefox worked perfectly. I still got a tiny difference for Chrome and Safari. Testing with the plyr demo videos maybe 1/10 time you click you get a second difference between the tooltip and seek time. Testing with a 2 hour movie on youtube it's as much as 10 seconds (instead of up to 2 minutes).

The second commit fixes this by adding an alternative `value` calculated with the exact same way as the tooltip times, before the seek slider's change event.

Users can still change the thumb size by changing the sass variable or custom css to override every place it's used. Not optimal, but at least possible :/

### DIFF
--- a/src/sass/components/progress.scss
+++ b/src/sass/components/progress.scss
@@ -6,10 +6,15 @@
     display: flex;
     flex: 1;
     position: relative;
+    margin-right: $plyr-range-thumb-height;
+    left: $plyr-range-thumb-height / 2;
 
     input[type='range'] {
         position: relative;
         z-index: 2;
+        // Offset the range thumb in order to be able to calculate the relative progress (#954)
+        width: calc(100% + #{$plyr-range-thumb-height}) !important;
+        margin: 0 -#{$plyr-range-thumb-height / 2} !important;
     }
 
     // Seek tooltip to show time
--- a/src/js/controls.js
+++ b/src/js/controls.js
@@ -443,7 +443,7 @@ const controls = {
 
         // Calculate percentage
         let percent = 0;
-        const clientRect = this.elements.inputs.seek.getBoundingClientRect();
+        const clientRect = this.elements.progress.getBoundingClientRect();
         const visible = `${this.config.classNames.tooltip}--visible`;
 
         const toggle = toggle => {
--- a/src/js/listeners.js
+++ b/src/js/listeners.js
@@ -529,12 +529,30 @@ class Listeners {
             }
         });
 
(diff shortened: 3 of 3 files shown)


friday:
An alternative could be to add a hidden slider with the same size and trigger click with coordinates when hovering the visible slider. I never tried this, but from what I could read it should be possible. This is even uglier if you ask me, but perhaps the result is more exact. Screen readers might not like it though.

The opposite: Making both events click events would also allow a consistent way to calculate the time, but this is less accessible and would make the progress bar jump.

It's also possible to make the handle really small, and then use `transform: scale();` to compensate along with some other changes to center it etc. This doesn't fix the issue however. Just makes the diff smaller, even if it's smaller than a pixel (and it needs to be at least 2px or border-radius will fail).

src/sass/components/progress.scss
```diff
@@ -6,10 +6,15 @@
...
     flex: 1;
     position: relative;
+    margin-right: $plyr-range-thumb-height;
+    left: $plyr-range-thumb-height / 2;
 
     input[type='range'] {
         position: relative;
         z-index: 2;
+        // Offset the range thumb in order to be able to calculate the relative progress (#954)
+        width: calc(100% + #{$plyr-range-thumb-height}) !important;
```
sampotts: Is the `!important` required for these two?

friday: It's needed to overspecify `.plyr--full-ui input[type='range']` (sliders.scss), but it's not the only way.

Changing the selector on line 5 to `.plyr--full-ui .plyr__progress` (actually did this first but changed it since it was less self-explanatory) works too for example, and changing the order of the imports might work as well.

sampotts:
The only bit with this one that concerns me is the overriding of the standard behaviour of the slider. I'm not sure how this will play with libs like range touch (another one of mine) that help with touch devices. They don't work too great with `<input type="range">` or at least iOS doesn't. I'll give it a test.

friday:
You are right. I did some testing and dragging the handle is "laggier" or even breaks. Clicking works.

friday:
I changed it so the place it would previously prevent the default action, instead it sets a new property `seekNext`, which when present is used instead of `value` for the next `input` or `change` event.

Issue #921, opened 2018-04-26
Youtube: Seeked event fired only when video is playing


lazic-st:
As title says, seeked event is fired only if video is playing.
This could be issue only with YouTube, I didn't test with any other video source. I understand that YouTube is tricky because they have really retarded API for player... (only one worthwhile event)

### Expected behaviour
Seeked event is event that should be fired after mouse keyup event on 'seeking pin'.

### Actual behaviour
Seeked event is fired everytime seeking event happen followed by play event. Which means, if player is in PAUSED state seeked event is not fired.

### Environment

- Browser: Chrome
- Version: latest
- Operating System:  Windows
- Version: 10

### Steps to reproduce
- Embed any YouTube video with Plyr
- add event listener with console.log on event seeked()
- open page, play and pause video
- try to seek video

gurupras:
I'm able to reproduce the issue reported here. I have also confirmed that it only occurs on YouTube. Here's a [pen](https://codepen.io/gurupras/pen/NMwPyR) to reproduce this issue.

sampotts:
It’s likely due to their API not firing an event but I’ll take a look and see if I can mimic the event in lieu of the missing event.

Issue #876, opened 2018-04-06
Setting currentTime unexpectedly triggers play for YouTube and Vimeo


gurupras:
### Expected behaviour
Setting `currentTime` directly or through `forward()` should only seek the video and not change the state of the player
### Actual behaviour
On a video that has never been played, `plyr.currentTime = 30` will seek 30 seconds in and start playing the video. On a paused video, this will only seek
### Environment
- Browser: Chrome/Firefox
- Version:
  - Chrome: 65.0.3325.181
  - Firefox: 59.0.1
- Operating System: Windows 10
- Version: Build 1703

### Steps to reproduce
Here's a [pen](https://codepen.io/gurupras/pen/LdgeeO)

gurupras:
This does seem to be the default HTML5 video behavior. Unfortunately, attempting to pause on `seeked` doesn't seem to work

```
plyr.on('seeked', function _once () {
  plyr.off('seeked', _once)
  if (expectedState === 'paused') {
    plyr.pause()
  }
})
```

gurupras:
Yeah, you're right. I think it may have been some listener I had active without realizing. I tried reproducing this on the [w3org-mediaevents](https://www.w3.org/2010/05/video/mediaevents.html) page and was unsuccessful.

The core problem pointed out in this issue still holds though.  
Should I just handle this manually rather than expect `plyr` to do it for me? I need to seek to a specific part of the video while also ensuring that the video state is respected. To do this, I'm currently having to do the following:
```
plyr.on('canplay', function () {
  if (startTime !== 0) {
    // Changing currentTime seems to implicitly trigger play
    plyr.on('seeked', function _once () {
      plyr.off('seeked', _once)
      if (expectedState === 'paused') {
        // Directly calling plyr.pause() does not work for some reason
        setTimeout(() => { plyr.pause() }, 10)
      }
      plyr.currentTime = startTime
    }
  })
})
```

sampotts:
I can't reproduce this on either Chrome or Firefox as stated.

As a side note there's a `playing` getter on the `Plyr`.

sampotts:
OK just seen it's with YouTube.

gurupras:
I thought I did follow the issue template, but yes, I should've made it a lot more clearer that the issue was with youtube/vimeo and not native HTML videos. The fact that I had an unreproducible follow-up comment  only made it worse. Sorry about that.

sampotts:
[...]

PR #959, opened 2018-05-12
Youtube and vimeo fixes

friday:
I merged these into one PR since I couldn't submit them separately without conflicts, but they are separate commits.

#### #876 Fix Youtube and vimeo autoplay on initial seek

Prevents the default behavior to start playing when seeking for the first time, by muting, pausing and then unmuting again.

Uses the youtube and vimeo apis to change state, but never sets the state to the instance, in order to restore it from the instance and apply.

Seeking while paused for youtube is still pretty broken, but this is a separate issue.

####  #921: Trigger seeked event in youtube plugin if either playing or paused

Before it was only triggered in the "playing" part of the switch statement.

#### #966: Fix youtube seek
Added `seeked` event listeners since currentTime isn't right for the `seeking` event (should fix that separately). Fixees seek for youtube (broken before, but you couldn't tell unless using it while paused)

#### Fix playback state (paused) and events (play/pause)
The `play` and `pause` event means that the `paused` state has changed (w3c words it like that, mdn didn't, but does now since I updated the event page). You can test at: http://jsbin.com/jutiveyefa/edit?console,output
Also made sure the events were triggered both in the `play` and `pause` methods (as they should be) or when detecting that the state changes via the API methods.

#### Prevent durationchange events from showing time when invertTime is false
If invertTime is false the currentTime should be shown when seeking or playing, but since youtube (and perhaps vimeo) can get the duration via the `durationchange` after playing/ initial seek, this causes the remaining time to show instead. 

I'm not sure this is the proper place to fix it.

### DIFF
--- a/src/js/plugins/youtube.js
+++ b/src/js/plugins/youtube.js
@@ -424,6 +424,17 @@ const youtube = {
                     // Reset timer
                     clearInterval(player.timers.playing);
 
+                    const seeked = player.media.seeking && [
+                        1,
+                        2,
+                    ].includes(event.data);
+
+                    if (seeked) {
+                        // Unset seeking and fire seeked event
+                        player.media.seeking = false;
+                        utils.dispatchEvent.call(player, player.media, 'seeked');
+                    }
+
                     // Handle events
                     // -1   Unstarted
                     // 0    Ended
@@ -457,12 +468,6 @@ const youtube = {
                             break;
 
                         case 1:
-                            // If we were seeking, fire seeked event
-                            if (player.media.seeking) {
-                                player.media.seeking = false;
-                                utils.dispatchEvent.call(player, player.media, 'seeked');
-                            }
-
                             // Restore paused state (YouTube starts playing on seek if the video hasn't been played yet)
                             if (player.media.paused) {
                                 player.media.pause();
--- a/src/js/plugins/youtube.js
+++ b/src/js/plugins/youtube.js
@@ -295,22 +295,17 @@ const youtube = {
                             return Number(instance.getCurrentTime());
                         },
                         set(time) {
-                            // Vimeo will automatically play on seek
-                            const { paused } = player.media;
+                            // If paused, mute audio preventively (YouTube starts playing on seek if the video hasn't been played yet).
+                            if (player.paused) {
+                                player.embed.mute();
+                            }
 
-                            // Set seeking flag
+                            // Set seeking state and trigger event
                             player.media.seeking = true;
-
-                            // Trigger seeking
                             utils.dispatchEvent.call(player, player.media, 'seeking');
 
                             // Seek after events sent
                             instance.seekTo(time);
-
-                            // Restore pause state
-                            if (paused) {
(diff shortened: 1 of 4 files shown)


sampotts:
Should we update on both `seeking` and `seeked`? Otherwise the scrubber would indicate a different time to the actual time displayed?

src/js/plugins/youtube.js
```diff
@@ -273,22 +274,19 @@ const youtube = {
                             return Number(instance.getCurrentTime());
                         },
                         set(time) {
-                            // Vimeo will automatically play on seek
-                            const { paused } = player.media;
-
                             // Set seeking flag
                             player.media.seeking = true;
 
+                            // Youtube will automatically play on seek if the video hasn't been manually paused, so mute and unpause later
```
sampotts: Minor typo but it's YouTube. A pet hate of mine with WordPress too haha

friday:
> Should we update on both seeking and seeked? Otherwise the scrubber would indicate a different time to the actual time displayed?

I really don't understand the reason for the separate update events and methods to begin with (as opposed to all the update methods listening to currentTime and update accordingly), so I primarily changed the parts I needed to make youtube work, but I may have broken the scrubber. I don't know what a scrubber is, but just noticed the initial time is shown as 0:00. Is that it?

src/js/plugins/youtube.js
friday: Sure thing!

sampotts:
Ah sorry, "scrubber" is like an industry term some folks seem to use for the seek input. 

I'm not sure what you mean by the other part. I tried to keep the methods small to update things like the buffer display, times etc and then only binding to the events required for those methods. Are you suggesting we have one mega method to handle it all? My preference is usually smaller methods with limited responsibility. I'm sure there could be cleanup though. Also, you can't really bind to currentTime (perhaps some sort of mutation observer?), there needs to be an event to listen to to know about the update - in the case of a change to currentTime, the event triggered is `timeupdate`. 

I'll merge this and see but really the time should update to the intended seek time. It may still do this but I'm not 100% sure.

friday:
Yes, `timeupdate` of course, not currentTime. 👍 

You may have good reasons for it, but it was a bit confusing to see the intention behind it. If you test the events you'll see some are triggered three times per change in vimeo and `timeupdate` isn't triggered for youtube. I only noticed after changing `seeking` to `seeked` which made `timeupdate` internally redundant (and then actually forgot about it).

I see why you wouldn't want circular logic or updates that interferes with user interactions though.

> I'll merge this and see but really the time should update to the intended seek time. It may still do this but I'm not 100% sure.

It doesn't (only for html5 media). You need to set currentTime in the player.currentTime setter synchronously as well as the async API calls. I think this is a good idea too. And then 143c821 wouldn't be needed for youtube seeking, so you can revert/skip it.

friday:
> Should we update on both seeking and seeked?

Forgot to communicate that I changed this. It now does that. Except only the `seeked` event actually works since the time doesn't update in `seeking`.

PR #961, opened 2018-05-14
Enable overriding defaults

friday:
Enables changing the defaults to whatever you use the most. For example:

`Object.assign(Plyr.defaults, {invertTime: false, ratio: '16:9', clickToPlay: false, controls: ['play', 'progress', 'current-time', 'mute', 'volume']});`

Rather than having to pass those options each time you initiate a Plyr instance.

I'm quite used to this from other libraries and it would simplify my own Plyr usage as well since the webapp I'm working for uses Plyr in five different places.

The only actual code needed to support this is `Plyr.defaults = defaults`, but that would introduce a risk of users accidentally removing defaults. Ex `Object.assign(Plyr.defaults, {tooltips: {controls: true}})` would remove `defaults.controls.seek`.

I didn't document the changes. Not sure which section to add it to (or if you even like the idea).

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -55,6 +55,7 @@ class Plyr {
         this.config = utils.extend(
             {},
             defaults,
+            Plyr.defaults,
             options || {},
             (() => {
                 try {
@@ -1268,5 +1269,7 @@ class Plyr {
         return targets.map(t => new Plyr(t, options));
     }
 }
+
+Plyr.defaults = utils.cloneDeep(defaults);
 
 export default Plyr;
--- a/src/js/utils.js
+++ b/src/js/utils.js
@@ -703,6 +703,11 @@ const utils = {
         return array.filter((item, index) => array.indexOf(item) === index);
     },
 
+    // Clone nested objects
+    cloneDeep(object) {
+        return JSON.parse(JSON.stringify(object));
+    },
+
     // Get the closest value in an array
     closest(array, value) {
         if (!utils.is.array(array) || !array.length) {


sampotts:
I think it'd be a limited use case but worth doing I suppose. 

With the cloneDeep, is that faster than just doing `Object.assign({}, object)`? Using Object.assign seems cleaner at least.

Also, can `Plyr.defaults = utils.cloneDeep(defaults);` not be within the class itself? As a getter/setter or static perhaps?

sampotts:
You're right about `Object.assign`. I thought it actually cloned if you passed an empty object as the first argument like jQuery's `$.extend()`. Sadly not.  

image

> Static properties unfortunately doesn't exists in javascript classes (at least yet). The functional equivalent to static methods for properties is what I just did.

Actually they do. That article seems to be about public vs private. We already use static classes:
https://github.com/sampotts/plyr/blob/master/src/js/plyr.js#L1235

friday:
> Actually they do. That article seems to be about public vs private. We already use static classes:
https://github.com/sampotts/plyr/blob/master/src/js/plyr.js#L1235

The static keyword is only for methods. For other properties. you have to add them to the class later.

The class field declarations-proposal is for both static and private "fields". So you could avoid declarations in constructors (`this.x = false;`) as well as having to set static properties to the class after it's declared (what we're currently talking about).

Also see https://github.com/amiller-gh/es2015-default-class-properties

sampotts:
Ah totally misread then! Looks like you've done more homework than me 👍 

The private class properties will definitely be useful when they get introduced. It's a bit of a mess currently.

friday:
> The private class properties will definitely be useful when they get introduced. It's a bit of a mess currently.

Yeah, in particular when it's the only thing you need the constructor for. I always hated that.

PR #963, opened 2018-05-15
Make sure poster element isn't shown if the image isn't loaded

friday:
Makes sure the poster element is only shown if the poster url is valid (like native html5 posters). Otherwise it would just be a black element covering the video (which may or may not be black too depending on the first frame), or in the youtube case it would show the fallback thumbnail, which is worse imo.

Regarding youtube: "sd" really is higher quality than "hq", unlike the name suggests. So this is not a mistake.

Both those formats are padded to 4:3. Youtube solves this by using `background-size: cover`, so I did that too (only for those).

I tried to "unpad" with css scale first. It works if you know the original aspect ratio, but since it's not possible to get that before playing the video "cover" seems good enough. "100% auto" should work better for landscape mode (makes it letterboxed as needed rather than "zoom" it), but not for portrait.

### DIFF
--- a/src/js/defaults.js
+++ b/src/js/defaults.js
@@ -199,7 +199,6 @@ const defaults = {
         youtube: {
             sdk: 'https://www.youtube.com/iframe_api',
             api: 'https://www.googleapis.com/youtube/v3/videos?id={0}&key={1}&fields=items(snippet(title))&part=snippet',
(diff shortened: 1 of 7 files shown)


sampotts:
Nice work. Do we need to reset/remove the `background-size` when changing source?

friday:
Good idea. I forgot about that. Will add that and rebase the last commit.

friday:
Had to make some more changes to `setPoster`, mainly to prevent race conditions with setting and resetting `background-size`. The promise also returns the poster url now, not the image created by utils.loadImage.

PR #967, opened 2018-05-17
toggleControls rewrite

### LINKED ISSUE #883: Controls are not showing when calling .toggleControls(true)
### Steps to reproduce
- Go to https://plyr.io/
- Play the video
- Type `player.toggleControls(true)` into the console

### Expected behaviour
Controls are shown when I call `player.toggleControls(true)`.

### Actual behaviour
Nothing happens when I call `player.toggleControls(true)`.

### Environment
- Browser: Chrome
- Version: 65
- Operating System: macOS
- Version: 10.12.6

friday:
Fixes #883.

* Makes `utils.toggleClass()` work like `Element.classList.toggle()`, by allowing the toggle argument (renamed to `force` to match `Element.classList.toggle` documentation) to be optional.
* Removes unused method ui.checkFailed() and related code, since this code gave the allure of integrating with toggleControls.
* toggleControls is split up in two methods.
  * The one on the plyr instance is "dumb" and ignores any state as long as the ui is supported and it's not audio. It also triggers the `controlshidden` and `controlsshown` event. Like `utils.toggleClass`, the actual toggle argument is optional.
  * `ui.toggleControls` toggles the controls based on state (`loading`, `paused`, `controls.pressed` and`controls.hover`). If any of those are true, or if you pass true as the `force` argument the controls will be shown. Otherwise not
  * All the event handling code is in listeners.js, and instead of deep nested conditions it checks and sets state in order to have `ui.toggleControls` handle it.

I removed `mouseenter` and `touchend` since `mousemove` / `touchmove`, handles these cases. At least in my tests in Chrome (with/without "device" simulation) `mouseenter` is always followed by `mousemove` and `touchend` should only happen after the hide timeout is already triggered by `touchstart` or `touchmove`.

The `noTransition` class is applied on `focusin` and removed on `focusout` regardless of the device being a touch device or not. Previously this was only for non-touch devices, but since touch devices can have keyboards this wasn't a safe assumption I think. I assumed it was added to prevent the browsers from forcing the focused element into view prematurely (causing the parent to scroll and look broken). This issue can be replicate on the plyr demo with the device simulator and focusing a button into view via keyboard.

There is an issue with the youtube object not setting `paused` to false before triggering the `play` event that I'll try to fix separately in the youtube/vimeo PR to avoid conflicts.


sampotts:
Nice work. I guess we should put some of this stuff in a branch other than master so I can test? 

The error stuff was WIP but I guess I can come back to this PR to grab that again.

friday:
You can test a PR locally without merging. I use [git-extras](https://github.com/tj/git-extras) (recommended since it's very easy to switch and not much to learn) which has an alias [`git pr`](https://github.com/tj/git-extras/blob/master/Commands.md#git-pr) for it [or this way](https://coderwall.com/p/_8dxjw/merging-a-pull-request-locally) without git extras.

I'm still working on the youtube/vimeo one since the events still aren't triggered in the right order with that PR as it is now. The `paused` property should always be set before the `play` and `pause` events are triggered for example. Now it's set after. Should be done tonight though.

Issue #980, opened 2018-05-28
Caption will show the first/last language as selected if the language isn't found in the tracks


friday:
If there isn't a corresponding track for the language for captions ex `new Plyr(element, {captions: {language: 'fi'}});` then the first or last track will falsely show as selected.

Can be reproduced this way:
* Go to https://plyr.io/#vimeo
* Select "Spanish". (Spanish is now stored as your localStorage language)
* Go to the HTML5 video.
* Click Settings: Will state English as language (since it's the first language)
* Click on captions: Will show French as the checked language (since it's the last language) 
* `player.language` will still be `es` if you try it in the console (but Spanish doesn't have a track).

Browsers: Replicated in Chrome and Firefox, osx

![screen shot 2018-05-28 at 21 47 05](https://user-images.githubusercontent.com/515120/40627697-beb4c758-62c0-11e8-9a72-1e86dcac6f4a.png)

![screen shot 2018-05-28 at 06 49 29](https://user-images.githubusercontent.com/515120/40597699-519a9f1c-6243-11e8-96bd-7465a5218106.png)

![screen shot 2018-05-28 at 21 47 51](https://user-images.githubusercontent.com/515120/40627706-ce5d32da-62c0-11e8-9c9d-c9002a2348a2.png)

Issue #971, opened 2018-05-22
Captions on hls.js


indiecore-fm:
I noticed that plyr supports hls.js. This is great! Now looking at hls.js I see that they support 608 captioning. 

I run a deaf video platform and we are looking for a open source video player to get out of using JW. 

Anyone have any guidance on how to use the caption features from hls.js with plyr? On the actual hls.js example it loads and plays the captions automatically for my stream.

I see the example here: https://codepen.io/sampotts/pen/JKEMqB But don't see how to add captions in.

friday:
You can add vtt captions as html5 tracks (which hls.js ignores) for Plyr. That's your only option for now.

sampotts:
How do you normally set the captions with a HLS stream? We could easily add a setter to set the captions but I'm unfamiliar with the format they are usually set using HLS.

friday:
You get the current index:
```
hls.subtitleTrack
```
(will be `-1` if no track is selected)

You get the list via:

```
hls.subtitleTracks
```

Which will output something like:

```
[
  {
    "groupId": "subtitles",
    "name": "Swedish",
    "type": "SUBTITLES",
    "default": false,
    "autoselect": false,
    "forced": false,
    "url": "sv.m3u8",
    "lang": "sv",
    "id": 0,
    "details": {
      "type": "VOD",
      "version": 3,
      "url": "sv.m3u8",
      "fragments": [
        {
          "_url": "sv.vtt",
          "_byteRange": null,
          "_decryptdata": {
            "method": null,
            "key": null,
            "iv": null,
            "_uri": null
          },
          "tagList": [
            [
              "INF",
              "86400.0"
            ]
          ],
          "_elementaryStreams": {
            "audio": false,
            "video": false
          },
          "duration": 86400,
          "title": null,
          "type": "subtitle",
          "start": 0,
          "levelkey": {
            "method": null,
            "key": null,
            "iv": null,
            "_uri": null
          },
          "sn": 1,
          "level": 0,
          "cc": 0,
          "baseurl": "sv.m3u8",
          "relurl": "sv.vtt",
          "trackId": 0,
          "loaded": 55
        }
      ],
      "live": false,
      "startSN": 1,
      "targetduration": 86400,
      "totalduration": 86400,
      "averagetargetduration": 86400,
      "endSN": 1,
      "startCC": 0,
      "endCC": 0,
      "tload": 4991.900000022724
    }
  },
  {
    "groupId": "subtitles",
    "name": "English",
[...]

PR #981, opened 2018-05-28
Improve captions handling for streaming

friday:
Relates to #971

* Adds option to watch caption track changes and update language options
* Ensure languagechange and captionsenabled/captionsdisabled events are sent on initial setup, and that captionsenabled/captionsdisabled is only triggered on change.
* Ensure language is set in case the track is added after initialization
* Optimizations for code reuse.

`instance.toggleCaptions` will now ensure the elements are updated with the right classes/state even if they already were before. The reason for this is code reuse. I was able to remove the`captions.show` method this way. Without the changes to `instance.toggleCaptions` however it would not work if you load a new source in the same instance.

Another thing that may need explaining is `this.language` is being set in the new `captions.update` method (when it hasn't been set already). This is needed because setting `this.language` will rightfully bail if the language doesn't have a track (and log a warning which may need to be upgraded now). The track for the language may be added after initializing Plyr, so it has to try on every update until it has a match.

How to test:
```js
new Plyr(video, {captions: {update: true}});
video.addTextTrack('subtitles', 'Volapük', 'vo');
```

Hls.js introduces another dimension of issues to the table, and it's not really ours to fix, but if you want to test it, here's a new simplified jsbin (using this branch) with Plyr + hls.js and two kinds of captions:
http://jsbin.com/mayikihosu/edit?js,output

Not in this PR:
* It doesn't fix the css issue (I think?) causing #980, however it does ensure `player.language` isn't something that doesn't have a track. So the last bullet would be `undefined` not `es`
* Keeping an internal reference of the current track index - This would be nice for other streaming libraries as well since it seems to be how they operate (the same applies to quality options). The method I'm running in the `languagechange` listener in the jsfiddle examples is not safe without this, if there are duplicate tracks per language (like hard-of-hearing subtitles).


readme.md
```diff
@@ -303,7 +303,7 @@ Note the single quotes encapsulating the JSON and double quotes on the object ke
...
+| `captions`           | Object                     | `{ active: false, language: browser language, update: false  }`                                                         | `active`: Toggles if captions should be active by default. `language`: Sets the default language to load (if available). `update`: Listen to changes to tracks and update menu. This is needed for some streaming libraries, but can result in unselectable language options).                                                                                                                                                                                                                                               |
```
sampotts: How do you feel about perhaps making the default language `'auto'` and we translate that into `navigator.language` in the code? Might be more user friendly.

src/js/ui.js
```diff
@@ -55,8 +55,10 @@ const ui = {
         // Remove native controls
         ui.toggleNativeControls.call(this);
 
-        // Captions
-        captions.setup.call(this);
+        // Setup captions for html5
```
sampotts: Super minor but can we make this `HTML5`

readme.md
friday: I have actually already done that :), but that was for the SSR issue, and I ended up pushing a different solution. I agree this is better.

friday:
I changed some CORS settings for our test server so https://testing.studiotogo.com/project/59db9312345a8234c27c809b/playlist.m3u8 can be used to test multiple languages and VTT (the other one is embedded).

(you can replace the url in the jsfiddle examples).

friday:
I changed this to WIP since I found another issue (`languagechange` event occasionally triggering before event listener).

I also updated the description to replace the hls.js stuff with a simple test. Hls.js introduces another dimension of issues to the table, and it's not really ours to fix.

sampotts:
Is this one ok to merge now? Awesome work btw 👍

friday:
Yes. I'm no longer working on it. I had to do some changes and more testing, but it seems solid now :)

PR #983, opened 2018-05-28
Remove cors attributes for local links in the demo html

friday:
Cors-attributes aren't needed for same-origin content, and creates problems if someone wants to run the demo by opening the file directly in the browser (not sure if that was an intended use case).

Svg still doesn't work for running directly in the browser though.

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -175,7 +175,7 @@
         crossorigin="anonymous"></script>
 
     <!-- Plyr core script -->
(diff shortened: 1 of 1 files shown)


sampotts:
This is required for Sentry logging - we could possibly add this as part of the gulp build but it needs to be there according to Sentry.

friday:
I see.

No, it's not important imo. I just tried to help this user, but I actually prefer to say "we don't support that" and then not having to test this case either.

PR #988, opened 2018-05-30
Translate quality badges and quality names

philipgiuliani:
This is my second approach to https://github.com/sampotts/plyr/pull/878.
I have extended i18n to allow nested translations with the dot notation. So you can call `i18n.t('qualityName.720', this.config)`. And with this change I added the ability to translate the badge and the quality name.

Result:
bildschirmfoto 2018-05-30 um 14 58 44

… another way to solve this issue would be to add a `label` to the `<source>` elements. That would be the same approach as `captions` is currently using. But this seems like a complicated change to me.


philipgiuliani:
… another way to solve this issue would be to add a `label` to the `<source>` elements. That would be the same approach as `captions` is currently using. But this seems like a complicated change to me.

philipgiuliani:
Thank's for the review. For some reason the `ESLint tweak` commit remained in my branch, but I guess this is not a problem.

friday:
You can remove commits, combine etc with an interactive rebase (ex `git rebase -i HEAD~9`), but do read up on the topic if you haven't used this before. I'm not sure it's a problem per se. It depends on Sam's merging strategy.

philipgiuliani:
@sampotts What do you think about this changes?

src/js/controls.js
```diff
@@ -708,15 +688,20 @@ const controls = {
...
-    // TODO: Localisation
     getLabel(setting, value) {
         switch (setting) {
             case 'speed':
                 return value === 1 ? i18n.get('normal', this.config) : `${value}&times;`;
 
             case 'quality':
                 if (utils.is.number(value)) {
-                    return `${value}p`;
+                    const qualityName = i18n.get(`qualityName.${value}`, this.config);
```
sampotts: Small one but could we call the variable `label` and maybe the property for i18n as `qualityLabel`?

philipgiuliani:
Sure! I've renamed the variable 😄

sampotts:
Thanks mate! Awesome solution btw. I'll deploy a new version shortly. I just need to work on a small UI bug I've noticed.

PR #1009, opened 2018-06-07
Contributing document and codepen demo updates

friday:
I thought the docs could use new updated demos for all formats, to help people get started with minimal implementations when they are testing out Plyr, want to report issues or ask for help.

The links are templates, rather than "pens", meaning they get forked directly by visiting the links. They use unpkg for cdn, since that allows specifying only the major version (v3) of Plyr. So they will use the latest version without needing to be updated (until v4).

Since they're on my codepen, you may want to move them over to yours so you are in control of them. I will be updating the Dash.js and Hls.js ones at a not too distant future though.

I left the previous codepen with custom buttons, since these are meant to get started quickly and that one also demonstrates the JS API usage.

I removed the Streaming section completely since it was redundant with both, but maybe it should be kept and reference back to the templates?

CONTRIBUTING.md is mostly inspired from other libraries, but adapted for Plyr. Since many of our users have shown a remarkable skill for neglecting the issue template, it's another way to reach them. First time they create an issue they should see this:

screen shot 2018-06-07 at 04 25 24

I improvised a lot and you may not agree, or I may have missed something. Feel free to change it.

The part about testing the demo locally is too complicated I think. we should add an npm script for local development that ensures that dependencies are installed, and run gulp (with watch) and a server.

There's some duplication between CONTRIBUTING.md and the FAQ, but I was thinking I would change the last FAQ question to basically link to CONTRIBUTING.md once/if this gets merged.

There's also some duplication is the issue template and PR templates, but I think this is good. CONTRIBUTING.md should be read once, and the templates every time.

I intentionally made master the base branch for this PR, since that way I could make references to CONTRIBUTING.md if this gets merged, before the next release.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,6 +25,12 @@ Describe the issue as detailed as possible, answering these questions:
   * [Hls.js integration](https://codepen.io/pen?template=oyLKQb)
   * [Shaka Player integration](https://codepen.io/pen?template=ZRpzZO)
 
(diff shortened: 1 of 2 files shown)


sampotts:
Thanks mate 👌

PR #1015, opened 2018-06-08
Captions rewrite (use index internally to support missing or duplicate languages)

### LINKED ISSUE #1003: Captions with formatting can produce nested spans
The `getCueAsHTML()` method converts cue markup to html. For example:

```xml
<v John Green>The human tongue is like <i.foreignphrase><lang ja>wasabi</lang></i>: it's very powerful, and should be used sparingly</v>
```

This will be converted to:

```html
<span title="John Green">The human tongue is like <i class="foreignphrase"><span lang="ja">wasabi</span></i>: it's very powerful, and should be used sparingly</span>
```

`<span>` can be added for metadata, but since Plyr uses span for styling the cue directly this causes issues.

screen shot 2018-06-05 at 14 58 16

(Opacity has been increased to better illustrate the issue)

It happens in both Chrome and Firefox (haven't tested it in any other browser).

### LINKED ISSUE #1014: Captions are enabled by default when setting the default language to something else than auto.
### Steps to reproduce
1. Configure the player with the following settings
```
captions: {
  active: false,
  language: "de"
}
```
2. Ensure that a track with the given language (de) is available.

### Actual behaviour
The captions are enabled, even if `active` is set to false.

### Expected behaviour
Captions shouldn't be enabled. When I click on the CC icon, the selected language should be DE.

---

If I uncomment the following line https://github.com/sampotts/plyr/blob/master/src/js/plyr.js#L882 it works, but the language selector doesn't enable the captions anymore.

friday:
Had to change A LOT for this one.

* The track index is now used internally, rather than language
  * Duplicate languages or missing srclang is now not a problem
  * Getter/setter for track index added `currentTrack`
    * Accepted values: Index in media.textTracks or `-1` to toggle captions off (this is how hls.js and dash.js works)
  * The `language` setter will set the "wanted" language.
  * The `language` getter will get the language from the "computed" track. Note that the computed track might not have a language, in which case it will be an empty string.
   * The computed track work like this:
     * Use currentTrack to get the track if it is set
     * Or:
       * Sort the tracks based on if they were default
       * Get the first matching language, if any
       * Fallback to the first non-match
    * The "wanted" language is needed in case there isn't any matching track initially. If tracks are added with dash.js, hls.js or custom code it can match later. So it tries to always show the best alternative. It doesn't fully make sense imo, and I'd rather toggle captions off if the language doesn't match, but then the captions toggle button wouldn't be able to do anything in that case, since there might not be anything to toggle to.
* Added a `WeakMap` to keep metadata about the tracks
  * Primarily used to store the `default` for tracks without setting it on the track. Default is not a valid property of the TextTrack element. They're only on the dom elements, which there is no safe way to match.
  * Also used to prevent adding duplicate event listeners for tracks
  * Also used instead of `.options.captions` for "freezing" the available languages when `captions.update` is `false`.
* Tracks will now be removed from the menu and cues will be reset if the track is removed from the textTracks list (happens if its element is removed from the dom). This will happen even if `captions.update` is false, in order to prevent "stuck" captions and unselectable language options.
* The setCue/setText methods were rewritten to one method `updateCues`. It takes an optional input argument (array of strings) or when omitted will use the current track instead. And it handles multiple simultaneous captions (which was the whole point).
  * Previous `setCue` implementation required an input argument, but also required that this argument was always the current track or an event with the current track as `target`. So it didn't really need an input argument at all. It could just get the current track.
  * The `cuechange` event was only called from `setCue`, not `setText` (which vimeo uses)
  * After simplifying and fixing those ^ I didn't think there was any need for the separate methods, so I merged them. 
  * The reason for the name change is I find "update" to be a better name than "set" for operations that primarily updates/syncs data from the state to the UI, while "set" makes more sense for operations that sets the state.
  * I needed to create a method `utils.getHTML()` for getting the html from the documentFragment that `cue.getCueAsHTML()` returns in order to be able to use `Array.map()` and `Array.join()` to create the captions.
* Fixed #1003: Formatted captions issue
* Fixed #1014: captions.active option not respected
* Changed `controls.createMenuItem()` to take an object instead of the list of arguments
   * It makes it a lot easier to read (see: https://eslint.org/docs/rules/max-params)
   * This way, the order of the argument doesn't matter, so optional arguments can be skipped instead of passing null

Example (Hls.js): https://codepen.io/anon/pen/wXgbvg?editors=1010


friday:
Rebased to remove a couple of previous commit as it turns out we didn't need (dash.js workaround) one and the other (moving the Vimeo event handling to the plugin) was causing trouble with `source()`. Also melded some commits.

Should be done now. Have tested thoroughly with Vimeo, streaming libraries and some different settings, but I don't think it's possible to cover all angles.

src/js/captions.js
```diff
@@ -69,105 +69,153 @@ const captions = {
             ({ active } = this.config.captions);
         }
 
-        // Set toggled state
-        this.toggleCaptions(active);
+        // Get language from storage, fallback to config
+        let language = this.storage.get('language') || this.config.captions.language;
+        if (language === 'auto') {
+            [ language ] = (navigator.language || navigator.userLanguage).split('-');
```
sampotts: Are we missing the `[0]` at the end of this to get the 'neutral' language? eg. `en-GB` to `en`.

friday: I'm not fully sure I understand the question, but it was intentionally replaced with array destructuring (the brackets surrounding `language`).

You probably know this, but this is how it works:

```js
const arr = [1, 2, 3]
const [first] = arr;
const alsoFirst = arr[0];
console.log({ first, alsoFirst }); // {first: 1, alsoFirst: 1}
```

So, like before it only gets the `en` part of `en-GB`

sampotts: Ahhh yes of course. I missed the destructuring (still getting used to the new syntax) 👍

PR #1020, opened 2018-06-10
Vimeo: Update playback state and assure events are triggered on load

### LINKED ISSUE #1016: "Playing" event not triggered when autoplaying Vimeo embed
On desktop, when initialising a Vimeo Plyr instance with autoplay: true (I have a placeholder img with the player only being initialised on click), it seems the "playing" event is not triggered and the "play" icon doesn't change to "pause". The video itself actually starts playing. The player needs to be clicked 2 times to be paused, with the expected functionality returning afterwards.

Thanks for your help!

friday:
Fixes #1016 

Tested in Desktop browsers Chrome (autoplays), Firefox (autoplays) and Safari (doesn't autoplay), and with autoplay enabled or disabled. It works as expected:
  * Events are triggered
  * The play button will show if it's not playing, otherwise the pause button will show

The exception is Safari. When autoplay fails Vimeo will still return `false`. `getPaused()` will still return `false`, as if it's playing. It will trigger `play`, `playing` and `pause`, so in the end the state will still be right.

### DIFF
--- a/src/js/plugins/vimeo.js
+++ b/src/js/plugins/vimeo.js
@@ -312,6 +312,14 @@ const vimeo = {
         });
 
         player.embed.on('loaded', () => {
+            // Assure state and events are updated on autoplay
+            player.embed.getPaused().then(paused => {
+                assurePlaybackState.call(player, !paused);
+                if (!paused) {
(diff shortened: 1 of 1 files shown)


sampotts:
Again, cheers 👍

PR #1021, opened 2018-06-11
Fix for YouTube and Vimeo pausing after seek

friday:
Limited the "YouTube and Vimeo autoplays on seek"-fix to only pause if the video has never been played (which is also the only time it's needed), due to a clash with d733454.

Also prevents temporary muting for Vimeo, but not YouTube. Preventing this shouldn't be needed I think, but changing this for Vimeo didn't adding any redundant code, while the same doesn't apply to YouTube.


sampotts:
Cheers 👍

PR #1023, opened 2018-06-11
Make utils static

friday:
This is an internal change only (no bug fix).

Some methods in `utils` were referencing either `utils` or `utils.is` using the `this` keyword. I think this should be avoided since then the methods can't be used statically, without binding or wrapping them.


friday:
Thanks @sampotts! I think individual functions is an even butter idea :), but I didn't look at the bigger picture.

sampotts:
@friday I'll give it a go as part of the a11y stuff I'm doing 👍

PR #1024, opened 2018-06-11
Event "detail" is lost in the synthetic event bubble/proxy

friday:
The problem can be tested on the demo:

```js
player.on('qualityrequested', event => console.log('Requested quality:', event.detail.quality));
```

Then when you change quality it should say for example: "Requested quality: 720". But the actual quality is missing.


sampotts:
Cheers 👍

PR #1027, opened 2018-06-12
Minor code improvements for quality switching

friday:
Minor code improvements (imo) for quality, and changes needed in order to get closer to supporting streaming libraries.

* More DRY code using array methods and key/value map instead of conditions and `switch` statement (no functional changes)
* Moved the `qualityrequested` event so it always triggers (if there are qualities to pick from), since plugins and libraries might want to listen to this event to handle it (needed for hls.js and such)
* Filter out unsupported mimetypes in getSources() instead of the quality setter (also needed for hls.js and such, but it should be better for html5 as well that these options don't appear in the menu and then can't be selected)
* Move uniqueness filter from getQualityOptions to setQualityMenu (since calling `setQualityMenu ()` via libraries with duplicates would otherwise add duplicates to the menu)

I recommend reviewing each commit separately since it's going to be easier to understand the changes that way.

### DIFF
--- a/src/js/controls.js
+++ b/src/js/controls.js
@@ -641,9 +641,10 @@ const controls = {
         const type = 'quality';
         const list = this.elements.settings.panes.quality.querySelector('ul');
 
-        // Set options if passed and filter based on config
+        // Set options if passed and filter based on uniqueness and config
         if (utils.is.array(options)) {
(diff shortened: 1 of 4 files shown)


sampotts:
Cheers 👍

PR #1034, opened 2018-06-14
Suggestion: Remove array newline rule

friday:
Removed the eslint rules that forces newlines within arrays, since the rules make some otherwise valid and practical code such as `Object.entries(([key, value]) => ...)`, `[x, y].includes(z)` and array destructuring much less practical.

`{"minItems": 2}` suggests the code samples should be permitted, but eslint still rejects them. Even with it though, I personally think `const preventDefault = [32, 37, 38, 39, 40, 48, 49, 50, 51, 52, 53, 54, 56, 57, 67, 70, 73, 75, 76, 77, 79];` is a lot more convenient than one line per keycode for example.

Also removed the current newlines in arrays. Running `prettier` with `--write` using the conf does this too actually, so unless this is only happening for me locally(?) the rules were conflicting with each other too.

Hopefully you agree. If not this can be rejected. There are no functional changes.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -32,9 +32,7 @@
                 "message": "Use local parameter instead."
             }
         ],
-        "no-param-reassign": [2, { "props": false }],
-        "array-bracket-newline": [2, { "minItems": 2 }],
-        "array-element-newline": [2, { "minItems": 2 }]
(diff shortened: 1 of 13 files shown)


sampotts:
Good call 👍

PR #1036, opened 2018-06-15
Captions fixes (again)

friday:
Fixes a caveat mentioned in the rewrite from last time, and some related issues including the issue with `captions.active` not being respected when using the `source` setter (as can be seen with Vimeo on the demo page).

The main problem was that the events listening to caption changes was setting this to localStorage, so when `source` is first soft-`destroy`ing the old dom, the tracks are all removed. The update handler would (rightfully) then toggle captions off, which in turn was (not rightfully) saved to localStorage. Then when the new source is created, the setting was read from localStorage and since it said `active` was `false`, so the new source would have captions toggled off.

This was solved by added a `passive` flag (default: `true`) to internal captions methods

* Used to avoid overriding user preferences, when the user didn't intend this. For example updating state because of the available tracks changing, or setting it initially.
* The public API (`.language = ...`, `.currentTrack = ...` and `.toggleCaptions()`) wraps the internal methods but forces passive to `false`. When using the public API, this is regarded as the user's intention, so user preferences (localStorage and the internal state representing user preferences) is updated.
* Captions will be toggled off regardless of `captions.active` when the requested language doesn't have a match, until either:
  * A matching track is added (when `captions.update` is `true`)
  * User picks a track / language by the public API or using the menu.
  * User toggles captions on using the button or `.toggleCaptions()`. The is the only time a language will be "forced" to show even if there isn't any matching the preferred language. The logic for this was improved to match the tracks with all browser languages, in case the user is multilingual.

Also fixed the `transitionend` event listener, which was recently broken, making the menus stay at a fixed height and width, and as a result adding tracks while the menu pane is open didn't work (the new option was "cropped").

### DIFF
--- a/src/js/captions.js
+++ b/src/js/captions.js
@@ -7,7 +7,7 @@ import controls from './controls';
 import i18n from './i18n';
 import support from './support';
 import browser from './utils/browser';
-import { createElement, emptyElement, getAttributesFromSelector, insertAfter, removeElement, toggleClass } from './utils/elements';
+import { createElement, emptyElement, getAttributesFromSelector, insertAfter, removeElement, toggleClass, toggleState } from './utils/elements';
 import { on, triggerEvent } from './utils/events';
 import fetch from './utils/fetch';
 import is from './utils/is';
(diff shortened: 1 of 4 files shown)


sampotts:
Awesome work yet again 👍

PR #1038, opened 2018-06-16
Small captions fixes

friday:
* Re-add lowercasing language input (and also fallback on `auto` if someone sets the language to an empty string via config).
* Fix issue with toggling not showing the new track and storing the new value when forcing a language.

Sorry these slipped my attention while testing.


sampotts:
Cheers 👍

Issue #1018, opened 2018-06-08
Custom poster for providers (YouTube and Vimeo)


friday:
Current behaviour(s):

* HTML5 video can use custom posters via the `poster` property.
* YouTube and Vimeo uses their own posters
* Setting the poster with `player.poster = ...` works for any video format (HTML5, YouTube or Vimeo), but it can't be done directly after creating the instance due to the internal handling being async (it has to verify the url works to prevent a black element). Setting it too soon leads to a race condition. Ex: http://jsfiddle.net/fgegfjp7/1/

Suggested improvements
* It should be possible to set the poster directly after creating the instance without worrying about any race conditions.
* It should be possible to set the poster in youtube and vimeo via a property in the options or on the element.

wvffle:
What's the status of this issue? Are you working on this? what needs to be done to close it?

joshangehr:
Also curious about this one. I've been using the [`ready`](https://github.com/sampotts/plyr#events) event which seems to be working, but a parameter would definitely be useful.

frafor1988:
Well it's been 4 years now, what's the status ^_^

PR #1039, opened 2018-06-16
Fix poster race conditions

friday:
Initial step for #1018

* Filter out `null` and `undefined` values in `elements.setAttributes()`, since they would otherwise be set as strings.
* Copy the `poster` property when replacing the media element in vimeo and youtube (which is also why the change above was made).
* Add `passive` option to internal `ui.setPoster()` method (like with captions).
* Fix race conditions: Make `ui.setPoster()` set the poster attribute synchronously, but apply it to the dom async, and invalidate if the actual update if the attribute has been changed.
* Also added a `ready` promise in events.js that is used to wrap setPoster so it doesn't trigger too soon.

I'm pretty sure this needs revisiting. I noticed some quirks, but not directly related to or introduced by this PR.

### DIFF
--- a/src/js/plugins/vimeo.js
+++ b/src/js/plugins/vimeo.js
@@ -119,8 +119,11 @@ const vimeo = {
         iframe.setAttribute('allowtransparency', '');
         iframe.setAttribute('allow', 'autoplay');
 
+        // Get poster, if already set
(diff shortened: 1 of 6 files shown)


sampotts:
Cheers 👍

PR #1040, opened 2018-06-16
Switches code optimizations

friday:
Code optimizations to replace switches when simpler more DRY alternatives like key/value object literals or normal conditions could be used instead (although if you as me, this is always the case).

I have tested all of them except the YouTube error handling, since I can't find any media that makes it trigger, but it should be working.

The `controls.createProgress()`-change (04b5cff) I only tested that it created the child with the correct innerText. I'm not sure what the function of it is. Is this for accessibility or can these be shown somehow?

I avoided bigger changes, in particular in controls.js, in order to avoid conflicts with your accessibility work.

### DIFF
--- a/src/js/controls.js
+++ b/src/js/controls.js
@@ -724,32 +724,27 @@ const controls = {
         let value = null;
         let list = container;
 
-        switch (setting) {
-            case 'captions':
-                value = this.currentTrack;
-                break;
+        if (setting === 'captions') {
+            value = this.currentTrack;
+        } else {
+            value = !is.empty(input) ? input : this[setting];
 
-            default:
-                value = !is.empty(input) ? input : this[setting];
+            // Get default
+            if (is.empty(value)) {
+                value = this.config[setting].default;
+            }
 
-                // Get default
-                if (is.empty(value)) {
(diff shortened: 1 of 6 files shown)


src/js/config/defaults.js
```diff
@@ -178,6 +178,8 @@ const defaults = {
         disabled: 'Disabled',
         enabled: 'Enabled',
         advertisement: 'Ad',
+        airplay: 'AirPlay',
+        pip: 'PIP',
```
sampotts: The reason I didn't include these is they are a branded product and an acronym rather than words that need translating. I know this doesn't force people to localise them but it implies they should if they are in the list.

friday: I see! I thought that was unintended, but it make sense now that you say it. I rebased that so it should work like before now (but with an object literal).

sampotts:
A little bit less readable to me in some spots but I'll get the hang of it 👍

friday:
With the destructuring you mean? I'm using it constantly so you're probably going to have to get used to it ;)

sampotts:
Ah, more the removal of switch/case

Sam

Issue #1048, opened 2018-06-18
Play button not changing to show pause-icon when video is playing


wagich:
### Expected behaviour
The play button shows a pause-icon when media is playing.

### Actual behaviour
The play button shows a play-icon regardless of media state.

### Environment
- Browser: Chrome 67, Edge 42, Firefox 60
- Operating System: Windows 10
- Version: 3.3.15

### Steps to reproduce
- Go to https://plyr.io/
- Click play
- Observe play button in lower left corner

wagich:
It seems to be a tegression introduced in 3.3.15, 3.3.14 works.

friday:
Yeah, it was actually a fix for another issue which I think is non-fatal https://github.com/sampotts/plyr/commit/ea4d91d2a011cdfc835619cb90a25074c039e777#diff-21a888c55f507a76f2a82a75d300d6f7 Stay on  3.3.14 for now.

Thanks for reporting!

sampotts:
This is fixed in v3.3.16 - Sorry @friday I missed your change but ended up writing exactly the same code 👍

mogita:
It seems this issue appears again in 3.4.0

sampotts:
Try the latest version

PR #1050, opened 2018-06-18
Fix play button

friday:
#1048 

I'm not actually sure where `this.elements.buttons.play` is set, but in my tests it's an array or null, not a nodelist.


sampotts:
Ah I missed this one but made exactly the same change anyhow. Cheers! 👍

Issue #1044, opened 2018-06-18
preload=none shows no effect


MichaGue:
Hello,

is it correct, that using the standard html 5 attribute `preload="none"` has no effect using plyr? I did the following test in firefox, edge and chrome. In both cases the attribute `preload="none"` is set. In the first one, plyr is disabled. While there is no poster an empty html 5 video player will be shown. In the second one plyr is enabled. Even the attribute `preload="none"` is set and there is also no poster available the first frame is visible.

Is there a way to stop preloading using plyr?

Thanks a lot.
Michael

```
<video class="js-player" preload="none" controls playsinline id="player1" data-plyr-config='{ "volume": 0.5, "resetOnEnd": true, "enabled": false }'>
	<source src="360p.mp4" type="video/mp4" size="360">
	<source src="720p.mp4" type="video/mp4" size="720">
	<source src="1080p.mp4" type="video/mp4" size="1080">
</video>
		
<video class="js-player" preload="none" controls playsinline id="player2" data-plyr-config='{ "volume": 0.5, "resetOnEnd": true, "enabled": true }'>
	<source src="360p.mp4" type="video/mp4" size="360">
	<source src="720p.mp4" type="video/mp4" size="720">
	<source src="1080p.mp4" type="video/mp4" size="1080">
</video>
```

liamtarpey:
Have the same issue, have a page where I have a lot of audio player instances and a network call is made for each one as soon as the page loads. Is there a way to only load the file upon clicking play?

liamtarpey:
Actually I take this back. `preload="none"` has no effect as mentioned, the audio loads the metadata needed on init and only loads the actual rest of the data upon clicking play it seems.

friday:
I think this is preventable. Have fixed it locally, but I just have to make sure it's safe.

PR #1051, opened 2018-06-19
Respect preload="none" when setting quality if the media hasn't been loaded some other way

friday:
See #1044.

This was happening because the source is set from the default or localStorage.

I also moved the `loadedmetadata` event listener to the condition because I didn't want a latent event to trigger later and set currentTime to `0`. Not sure this would happen, but I felt uneasy leaving it.


sampotts:
Nice one 👍

Issue #1002, opened 2018-06-05
Max playbackRate option?


Astray-git:
Currently, the video playbackRate is limited to `2.0`.(Maybe because of Safari support)
Latest Chrome can play with sound at 16.0. Firefox can play with sound at 4.0.
Can you provide an option for max speed?

sampotts:
Hey,

I obviously need to document it better but you can specify the options here:
https://github.com/sampotts/plyr/blob/master/src/js/defaults.js#L92

So you can pass in more options in your config to override those defaults. 

Cheers 👍

Astray-git:
Thanks for the quick reply.
https://github.com/sampotts/plyr/blob/a4caba120c76a19fa9424598de119c44776e386d/src/js/plyr.js#L643-L645
I guess there's a 2.0 limit for `speed`

sampotts:
Ahh good spot. Well I can easily lift that. Will do.

sampotts:
In v3.5.7 you can now specify all the options you want. Options outside of 0.5-2x are not available in YouTube and Vimeo so they will get removed automatically.

PR #1053, opened 2018-06-19
Speed improvements

friday:
Relates to #1002, but also changes some other things.

I think I need some feedback on it since it seems to go against the intention behind the original API design.

Changes:
* `ui.setSpeedMenu()` was renamed `ui.updateSpeedMenu()` and now only updates the menu. Previously it also took an optional array and used it to set `this.options.speed` unless the provider was YouTube. But this option was only used from the Vimeo plugin for clearing the menu if `setPlaybackRate` failed. Instead the default `[0.5, 0.75, 1, 1.25, 1.5, 1.75, 2]` was used, and the config options was used to filter this array. So anything in the config which was not in this array was filtered out (also preventing #1002).
* `this.options.speed` is now set separately per provider. HTML5 uses the config directly. Vimeo uses the config, but filters out values it can't handle (under `0.5` or over `2.0`). YouTube ignores the config in favor of their own `instance.getAvailablePlaybackRates()`. Since the user config could be `[0.6, 0.7, 0.8, 0.9, 1, 1.1, 1.2, 1.3, 1.4, 1.5]` for example, filtering two arrays against each other could end up losing most or even all options. I think we have to improve our API (min/max rather than options), or even UI (slider perhaps?) to make sense of this.
* The `speed` setter now *only* uses the input. Previously it would fallback to the localStorage or config if you set it to a non-number. and also use min/max to contain the value between 0.1 and 2. I don't think this made sense. Now it instead gets the `closest` supported value from `this.options.speed`, so if you try to set it to `1.6` but this isn't supported you get the closest alternative ex `1.5`.
* Since the speed setter no longer "resets" if you pass a non-number. I had to change this in ui:

```js
this.speed = null;
```

to:

```js
this.speed = this.storage.get('speed') || this.config.speed.selected;
```

I don't think this is the right place for it though? I also needed a `setTimeout` to make this work, which felt like a hack. So pointers would be welcome.

### DIFF
--- a/src/js/controls.js
+++ b/src/js/controls.js
@@ -891,8 +891,8 @@ const controls = {
         controls.updateSetting.call(this, type, list);
     },
 
-    // Set a list of available captions languages
-    setSpeedMenu(options) {
+    // Update the list of available speeds
+    updateSpeedMenu() {
         // Do nothing if not selected
         if (!this.config.controls.includes('settings') || !this.config.settings.includes('speed')) {
             return;
@@ -904,16 +904,6 @@ const controls = {
         }
 
         const type = 'speed';
-
-        // Set the speed options
-        if (is.array(options)) {
-            this.options.speed = options;
-        } else if (this.isHTML5 || this.isVimeo) {
-            this.options.speed = [0.5, 0.75, 1, 1.25, 1.5, 1.75, 2];
-        }
-
-        // Set options if passed and filter based on config
-        this.options.speed = this.options.speed.filter(speed => this.config.speed.options.includes(speed));
 
         // Toggle the pane and tab
         const toggle = !is.empty(this.options.speed) && this.options.speed.length > 1;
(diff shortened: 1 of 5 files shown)


Astray-git:
Can we have an option for whether to save speed in storage?

sampotts:
> Can we have an option for whether to save speed in storage? 

Do you mean rather than disabling storage completely? I guess we could get more granular with it, yep.

friday:
I think that's a good separate suggestion.

Storage could be both an object or boolean, where `true` is translated to `{volume: true, muted: true, quality: true, speed: true, captions: true, language: true}` (and `false` to the opposite obviously).

sampotts:
@friday I agree with the change to the setter. Should we make this consist with the others though like quality, volume, muted, speed? They all seem to have similar behaviour (fall back to storage or config).

friday:
> Should we make this consist with the others though like quality, volume, muted, speed? They all seem to have similar behaviour (fall back to storage or config).

Yeah, I do think so. Setters shouldn't have fallbacks imo.

jmprovencher:
Any update on this ? Will it be merged?

spmeesseman:
i added to an issue that asked for more speed steps but just saw this, beautiful. will this be merged into main branch the future?

sampotts:
Closing this as I've cherry picked parts of it but left the setters logic for now. That probably needs a rework for all setters at the same time. Probably a `v4` thing.

PR #1061, opened 2018-06-21
Fix captions.toggle() if there is no toggle button

friday:
`.toggleCaptions()` (internally `captions.toggle()`) currently expects a toggle button.

It might not exist, in which case toggling should still be possible. This PR fixes that.


sampotts:
Cheers 👍

PR #1081, opened 2018-07-01
Improve flexibility of quality selector

gurupras:
I would like some feedback on this PR. If you use Plyr and you care about the quality setting, please have a read and provide feedback.

This is related to #878 and also PR #988

- All providers specify the following for each quality level
  - label: String _mandatory_ (This is what shows up in the menu. Solves #878)
  - height: Number _mandatory_ (Used to determine badge if badge isn't specified)
  - index: Number _optional_
  - badge: String _optional_

The 'index' field is optional and is used to resolve conflicts. An example of where this may come in handy is provided below.

Finally, the badge field allows the provider to explicitly specify a badge for a particular quality level. This is somewhat a double-edged sword. On one hand, this gives full freedom to providers to determine the badges for each quality level. On the other, it may lead to inconsistencies where one provider reports `HD` for a video with height of 720 whereas another provider does not.

Imagine a scenario where the available quality levels are as follows:
```
{width: 422, height: 180, bitrate: 258157}
{width: 638, height: 272, bitrate: 520929}
{width: 638, height: 272, bitrate: 831270}
{width: 958, height: 408, bitrate: 1144430}
{width: 1277, height: 554, bitrate: 1558322}
{width: 1921, height: 818, bitrate: 4149264}
{width: 1921, height: 818, bitrate: 6214307}
{width: 4096, height: 1744, bitrate: 10285391}
```

In the above example, with the current implementation, the only attribute that is passed on to the provider on `qualityrequested` is `value`. If the provider implemented this similar to how YouTube is currently implemented (purely based on height), then, there would be two quality entries with the same `value`. My hope is that the index field will help with resolution in such cases by telling the provider exactly which quality element was selected.

Even after #988 has been merged in, the quality badge relies heavily on some hard-coded values. If a video presents a quality value that is not found in [480, 576, 720, 1080, 1440, 2160], then no badge is provided. In the above example, _none_ of the entries would get a badge (if the `value` was determined by height).   
Instead, this PR allows providers to explicitly specify a badge of their choice. We could also add some fallback logic where the 'height' attribute is used to determine the badge.

Here's a [pen] that shows the results for HTML5/HLS/YouTube.
### Checklist
- [ ] Ensure PR makes sense with regard to `i18n`
- [ ] Figure out how to solve initial quality on HTML5 videos (currently shows `undefined`)

### DIFF
--- a/src/js/controls.js
+++ b/src/js/controls.js
@@ -373,7 +373,7 @@ const controls = {
     },
 
     // Create a settings menu item
-    createMenuItem({ value, list, type, title, badge = null, checked = false }) {
+    createMenuItem({ value, list, type, title, badge = null, checked = false, index = null }) {
         const item = createElement('li');
 
         const label = createElement('label', {
@@ -388,6 +388,7 @@ const controls = {
                 value,
                 checked,
                 class: 'plyr__sr-only',
+                index,
             }),
         );
 
@@ -671,6 +672,17 @@ const controls = {
     },
 
     // Set the quality menu
+    // options is expected to be an array of objects
+    // Each entry in this array should be of the type:
+    // {
+    //   label: String,      // mandatory
+    //   height: Number,     // mandatory
+    //   index: Number,      // optional
+    //   badge: String,      // optional
+    // }
+    // The order of qualities will be based on height. If there are multiple
+    // entries with the same height, then we use index instead.
+    // If badge is not specified, it will be looked up by height.
     setQualityMenu(options) {
         // Menu required
         if (!is.element(this.elements.settings.panes.quality)) {
@@ -680,9 +692,9 @@ const controls = {
         const type = 'quality';
         const list = this.elements.settings.panes.quality.querySelector('ul');
 
-        // Set options if passed and filter based on uniqueness and config
+        // Set options if passed
         if (is.array(options)) {
-            this.options.quality = dedupe(options).filter(quality => this.config.quality.options.includes(quality));
+            this.options.quality = dedupe(options);
         }
 
         // Toggle the pane and tab
@@ -702,10 +714,17 @@ const controls = {
 
         // Get the badge HTML for HD, 4K etc
         const getBadge = quality => {
-            const label = i18n.get(`qualityBadge.${quality}`, this.config);
+            let label;
+            if (quality.badge) {
+                label = quality.badge;
+            } else {
+                // The badge is based on the height of the video
+                // TODO: We need some rounding logic here
+                label = i18n.get(`qualityBadge.${quality.label}`, this.config);
 
-            if (!label.length) {
-                return null;
+                if (!label.length) {
+                    return null;
+                }
             }
 
             return controls.createBadge.call(this, label);
@@ -714,16 +733,19 @@ const controls = {
         // Sort options by the config and then render options
         this.options.quality
             .sort((a, b) => {
-                const sorting = this.config.quality.options;
-                return sorting.indexOf(a) > sorting.indexOf(b) ? 1 : -1;
+                if (a.height === b.height) {
+                    return a.index > b.index ? -1 : 1;
+                }
+                return a.height > b.height ? -1 : 1;
             })
             .forEach(quality => {
                 controls.createMenuItem.call(this, {
-                    value: quality,
+                    value: quality.label,
                     list,
                     type,
-                    title: controls.getLabel.call(this, 'quality', quality),
+                    title: controls.getLabel.call(this, 'quality', quality.height, toTitleCase(quality.label)),
                     badge: getBadge(quality),
+                    index: quality.index,
                 });
             });
 
@@ -731,7 +753,7 @@ const controls = {
     },
 
     // Translate a value into a nice label
-    getLabel(setting, value) {
+    getLabel(setting, value, defaultLabel) {
         switch (setting) {
             case 'speed':
                 return value === 1 ? i18n.get('normal', this.config) : `${value}&times;`;
@@ -740,14 +762,14 @@ const controls = {
                 if (is.number(value)) {
                     const label = i18n.get(`qualityLabel.${value}`, this.config);
 
+                    // If we don't find a valid label, we return passed in defaultLabel
                     if (!label.length) {
-                        return `${value}p`;
+                        return defaultLabel;
                     }
 
                     return label;
                 }
-
-                return toTitleCase(value);
+                return defaultLabel || toTitleCase(value);
 
             case 'captions':
                 return captions.getLabel.call(this);
@@ -773,16 +795,26 @@ const controls = {
                 value = this.config[setting].default;
             }
 
+            let settingOptions = this.options[setting];
+            if (setting === 'quality') {
+                // Quality is not an array of strings. It's an array of Objects
+                // Extract out the strings
+                settingOptions = this.options[setting].map(level => controls.getLabel.call(this, 'quality', level.label));
+            }
+
             // Unsupported value
-            if (!is.empty(this.options[setting]) && !this.options[setting].includes(value)) {
+            if (!is.empty(settingOptions) && !settingOptions.includes(value)) {
                 this.debug.warn(`Unsupported value of '${value}' for ${setting}`);
                 return;
             }
 
             // Disabled value
-            if (!this.config[setting].options.includes(value)) {
-                this.debug.warn(`Disabled value of '${value}' for ${setting}`);
-                return;
+            // Don't check for quality which is permissive and accepts anything
+            if (setting !== 'quality') {
+                if (!this.config[setting].options.includes(value)) {
+                    this.debug.warn(`Disabled value of '${value}' for ${setting}`);
+                    return;
+                }
             }
         }
 
--- a/src/js/controls.js
+++ b/src/js/controls.js
@@ -723,6 +723,7 @@ const controls = {
                 label = i18n.get(`qualityBadge.${quality.label}`, this.config);
 
                 if (!label.length) {
+                    // TODO: Try to find a badge based on height
                     return null;
                 }
             }
@@ -754,23 +755,19 @@ const controls = {
 
     // Translate a value into a nice label
     getLabel(setting, value, defaultLabel) {
+        let label;
         switch (setting) {
             case 'speed':
                 return value === 1 ? i18n.get('normal', this.config) : `${value}&times;`;
 
             case 'quality':
-                if (is.number(value)) {
-                    const label = i18n.get(`qualityLabel.${value}`, this.config);
+                label = i18n.get(`qualityLabel.${value}`, this.config);
 
-                    // If we don't find a valid label, we return passed in defaultLabel
-                    if (!label.length) {
-                        return defaultLabel;
-                    }
-
-                    return label;
+                // If we don't find a valid label, we return passed in defaultLabel
+                if (!label.length) {
+                    return defaultLabel || toTitleCase(value);
                 }
-                return defaultLabel || toTitleCase(value);
-
+                return label;
             case 'captions':
                 return captions.getLabel.call(this);
 
--- a/src/js/html5.js
+++ b/src/js/html5.js
@@ -21,10 +21,15 @@ const html5 = {
     // Get quality levels
     getQualityOptions() {
         // Get sizes from <source> elements
-        return html5.getSources
+        const sizes = html5.getSources
             .call(this)
             .map(source => Number(source.getAttribute('size')))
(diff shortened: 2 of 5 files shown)


gurupras:
@friday @philipgiuliani: Kindly take a look at this and provide feedback

sampotts:
PR should really be against develop

gurupras:
Ah, my bad. Updated.

src/js/plyr.js
```diff
@@ -677,21 +677,30 @@ class Plyr {
...
-            this.storage.get('quality'),
-            config.selected,
-            config.default,
-        ].find(is.number);
-
-        if (!options.includes(quality)) {
-            const value = closest(options, quality);
-            this.debug.warn(`Unsupported quality option: ${quality}, using ${value} instead`);
-            quality = value;
+        if (input instanceof String) {
```
sampotts: Could you use the type checkers we have? e.g. `is.string(input)`

src/js/plyr.js
```diff
@@ -677,21 +677,30 @@ class Plyr {
...
+            // We have only a label
+            // Convert this into an Object of the expected type
+            quality = {
+                value: input,
+            };
+        } else if (input instanceof Object) {
```
sampotts: Could you use the type checkers we have? e.g. `is.object(input)`

src/js/plugins/youtube.js
```diff
@@ -300,7 +307,13 @@ const youtube = {
                             return mapQualityUnit(instance.getPlaybackQuality());
                         },
                         set(input) {
-                            instance.setPlaybackQuality(mapQualityUnit(input));
+                            let label;
+                            if (input instanceof String) {
```
sampotts: Could you use the type checkers we have? e.g. `is.string(input)`

src/js/plugins/youtube.js
```diff
@@ -300,7 +307,13 @@ const youtube = {
...
+                                label = input;
+                            } else if (input instanceof Object) {
```
sampotts: Could you use the type checkers we have? e.g. `is.object(input)`

src/js/plugins/youtube.js
```diff
@@ -51,7 +51,14 @@ function mapQualityUnits(levels) {
         return levels;
     }
 
-    return dedupe(levels.map(level => mapQualityUnit(level)));
+    const mappedLevels = dedupe(levels.map(level => mapQualityUnit(level)));
+    const result = mappedLevels.map((level, index) => ({
```
sampotts: This can just be an early return since you're returning it on the next line anyway.

sampotts:
A few questions:

- Is this a breaking change? 
- Will the current options and markup work with this? 

I think having the fallback would be needed to allow the minimum effort. Not everyone will want to pass their own badge text. We can use the `closest()` array util or just something like `if (x > y) { badge = 'HD' }` to determine the best badge to display but allow devs to override these. That would be the ideal implementation I'd say?

gurupras:
I don't know how to be a 100% whether this is/isn't a breaking change.

Currently, on [plyr.io](https://plyr.io), `plyr.quality` returns a `Number`, this implementation returns a `String`.  In other words, this implementation returns the current quality `label`. This may count as a breaking change..

Also, at the moment, I just noticed that there seems to be a bug with the setter `plyr.quality = '1080'`, I'll fix this ASAP.

friday:
Thanks for this! I'll take a look at it soon. Can also check if it's breaking then. Can you fix the linting errors apart from the `no-cycle` ones (they were already there before)? They should be easy.

src/js/plugins/youtube.js
```diff
@@ -300,7 +306,13 @@ const youtube = {
                             return mapQualityUnit(instance.getPlaybackQuality());
                         },
                         set(input) {
-                            instance.setPlaybackQuality(mapQualityUnit(input));
+                            let label;
+                            if (is.string(input)) {
+                                label = input;
+                            } else if (is.object(input)) {
+                                label = input.label;
```
gurupras: Don't know how to fix this linting error.

friday: Destructuring is when you do this:
```js
const {label} = input;
```

Instead of this:
```js
const label = input.label;
```

Since it's already declared here you need to wrap with parenthesis:
```js
({label} = input);
```

src/js/controls.js
```diff
@@ -714,41 +734,40 @@ const controls = {
...
             })
             .forEach(quality => {
                 controls.createMenuItem.call(this, {
-                    value: quality,
+                    value: quality.label,
                     list,
                     type,
-                    title: controls.getLabel.call(this, 'quality', quality),
+                    title: controls.getLabel.call(this, 'quality', quality.height, toTitleCase(quality.label)),
                     badge: getBadge(quality),
```
friday: I think we should respect the label and its formatting as it was specified by the developers, and only call getLabel if they didn't pass a label.

This includes i18n and changing the casing. In most languages I know of title case is incorrect formatting, and it make no sense to give developers the option to translate their own input. They should specify it in the language they want it.

src/js/controls.js
```diff
@@ -714,41 +734,40 @@ const controls = {
...
                 });
             });
 
         controls.updateSetting.call(this, type, list);
     },
 
     // Translate a value into a nice label
-    getLabel(setting, value) {
+    getLabel(setting, value, defaultLabel) {
+        let label;
```
friday: I think you should restore getLabel (see above). Not sure why you removed `${value}p`?

src/js/html5.js
```diff
@@ -35,21 +40,26 @@ const html5 = {
...
+                if (!source) {
+                    return undefined;
+                }
+                const sourceIndex = sources.indexOf(source);
 
-                // Return size, if match is found
-                return source && Number(source.getAttribute('size'));
+                // Return label, if match is found
+                return qualityLevels[sourceIndex].label;
             },
```
friday: This should be `height`, not `label`

src/js/listeners.js
```diff
@@ -376,7 +376,7 @@ class Listeners {
         // Quality change
         on.call(this.player, this.player.media, 'qualitychange', event => {
             // Update UI
-            controls.updateSetting.call(this.player, 'quality', null, event.detail.quality);
+            controls.updateSetting.call(this.player, 'quality', null, event.detail.quality.label);
         });
```
friday: This should be `height`, not `label`

src/js/html5.js
```diff
@@ -35,21 +40,26 @@ const html5 = {
...
             set(input) {
                 // Get sources
                 const sources = html5.getSources.call(player);
 
                 // Get first match for requested size
-                const source = sources.find(source => Number(source.getAttribute('size')) === input);
+                const source = sources.find(source => source.getAttribute('size') === input.label);
```
friday: This should be `height`, not `label`

src/js/controls.js
gurupras: Is `272p` or `408p` legitimate? My understanding from your post on Slack was that the 'p' represents widescreen resolutions. Formally it's progressive scanning rather than interlaced.

src/js/controls.js
```diff
@@ -680,9 +692,9 @@ const controls = {
         const type = 'quality';
         const list = this.elements.settings.panes.quality.querySelector('ul');
 
-        // Set options if passed and filter based on uniqueness and config
+        // Set options if passed
         if (is.array(options)) {
-            this.options.quality = dedupe(options).filter(quality => this.config.quality.options.includes(quality));
+            this.options.quality = dedupe(options);
```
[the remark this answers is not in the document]
gurupras: Shall I replace this with just 
```
if (is.array(options) {
  this.options.quality = options
}
```

src/js/controls.js
friday: No that was `k` As in `2K` or `4K` video etc. `408p` basically means it's 408px height.

src/js/controls.js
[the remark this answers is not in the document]
friday: Yes! :)

src/js/plugins/youtube.js
```diff
@@ -51,7 +51,13 @@ function mapQualityUnits(levels) {
         return levels;
     }
 
-    return dedupe(levels.map(level => mapQualityUnit(level)));
+    const mappedLevels = dedupe(levels.map(level => mapQualityUnit(level)));
+    return mappedLevels.map((level, index) => ({
+        label: levels[index],
+        index,
+        height: level,
+        badge: level >= 720 ? 'HD' : 'SD',
```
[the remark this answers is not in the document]
gurupras: Removing badge in favor of default logic

src/js/controls.js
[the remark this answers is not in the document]
friday: (Sam might disagree but at least it won't look like it's doing something and then it doesn't work)

src/js/html5.js
```diff
@@ -21,10 +21,15 @@ const html5 = {
...
+        const sizes = html5.getSources
             .call(this)
             .map(source => Number(source.getAttribute('size')))
             .filter(Boolean);
+        return sizes.map((size, index) => ({
+            label: `${size}`,
+            height: size,
+            index,
+        }));
     },
```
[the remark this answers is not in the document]
gurupras: The template literal is to ensure that label conforms to the laid down rule (by me) that label is a `String`.

Updated with your suggestion. This is neat. I like being able to specify label, badge via `data-` attributes.

src/js/controls.js
```diff
@@ -702,10 +714,18 @@ const controls = {
...
+                // TODO: We need some rounding logic here
+                label = i18n.get(`qualityBadge.${quality.label}`, this.config);
 
-            if (!label.length) {
-                return null;
+                if (!label.length) {
+                    // TODO: Try to find a badge based on height
+                    return null;
+                }
             }
```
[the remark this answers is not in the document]
friday: I'm not sure what comment by Sam you're referring to, but you don't have to fix that. In fact, since it's a separate problem it's better not to.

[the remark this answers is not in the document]
gurupras: It's referenced [here](https://github.com/sampotts/plyr/pull/1081#issuecomment-401576060)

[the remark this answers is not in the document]
friday: Alright. I did read that, but too quickly I guess. I still think this should be a separate PR optimally. Solving it with closest also means 719p would count as HD, which it by definition isn't. I would expect it work like breakpoints, if anything. Sorry I'm making this hard for you btw!

gurupras:
No worries at all. I'd rather get it right than have others have to fix
issues I introduce later.

src/js/controls.js
```diff
@@ -714,41 +734,40 @@ const controls = {
         // Sort options by the config and then render options
         this.options.quality
             .sort((a, b) => {
-                const sorting = this.config.quality.options;
-                return sorting.indexOf(a) > sorting.indexOf(b) ? 1 : -1;
+                if (a.height === b.height) {
+                    return a.index > b.index ? -1 : 1;
+                }
+                return a.height > b.height ? -1 : 1;
             })
```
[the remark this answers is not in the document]
gurupras: I don't think that's the order you want.
```
["576", "1080", "720"].sort()
["1080", "576", "720"]
```
The previous implementation was actually doing it by height. At least on plyr.io.

sampotts:
Good point. It should be the second option then I guess. If larger or equal to 720 then it’s HD, larger than or equal to 1440 then it’s QHD etc. My point was I don’t think we need to force users to specify the labels. If they do it should override but if they don’t then we use sensible defaults to make life easy as it is now. 

Sam

src/js/plyr.js
```diff
@@ -677,21 +676,40 @@ class Plyr {
         const config = this.config.quality;
```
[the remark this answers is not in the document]
gurupras: Part of your comment, I think, is referring to the fact that this setter is trying to account for number/string/object all-in-one.

The number version was added just now to be backwards compatible. At least with plyr.io. It wasn't meant to handle index.

The string version is to cater to users wanting to set it by label.

The object version is mainly to handle the internal implementation wherein:
user clicks quality option --> translate `event.target.value` into quality somehow. This was where the object came into play.

Now, with the suggestion to remove index attribute and use it as the value while creating radio button, we can update the quality setter to receive either `["0", 0, { index: 0 }]`. I opted for the object implementation to allow sufficient differentiation.

gurupras:
I've incorporated most of the code-related feedback and made some fixes. Inviting more comments/feedback.

If required, I can add the default badge logic in this PR as well. I'm currently hesitant to make a separate PR for it as it would lead to some annoying merge conflicts if this PR were to be approved.

Issue #1085, opened 2018-07-04
iOS 9 Array.from requires an array-like object - not null or undefined


Maqsyo:
### Actual behaviour
on plyr.io:
![screen shot 2018-07-04 at 10 14 41](https://user-images.githubusercontent.com/19644698/42265447-dc24ae00-7f73-11e8-90e8-dff20194bbc2.png)

![image](https://user-images.githubusercontent.com/19644698/42265427-cd5e6186-7f73-11e8-8f77-84ceee7d7437.png)

with this error i'm unable to use it in my application.

Audios, vimeo and youtube are working fine. 
Videos fail

### Environment
- Plyr version: 3.3.21
- Browser: Safari for iOS
- Version: Default
- Operating System: iOS
- Version: 9

### Steps to reproduce
- It fails on page initialize

friday:
Would be great to get the Plyr line number.

Maqsyo:
On plyr.io it only shows the exception in the raven-file

In my app it seems to be a bit more detailed, so i could provide you this.
Please note that the line-numbers are incorrect.
![screen_shot_2018-07-06 at 10 53 22](https://user-images.githubusercontent.com/19644698/42369859-7bc17766-810b-11e8-892a-88c887afe402.png)

Mp4ct:
I think it's line 2894 in plyr.js. The problem is, that navigator.languages and navigator.userLanguage are both undefined in Safari on iOS 9. A fix could be to add || navigator.language.

Maqsyo:
The merge would fix it

Waiting for a raised version to use it :)

Maqsyo:
Please make a new release, so i could use this fix in productive projects @sampotts

sampotts:
Sure - I'll do a release in a few hours 👍

sampotts:
This is in v3.3.22

PR #1091, opened 2018-07-09
Add navigator.languages fallback for ios 9

friday:
Restore `navigator.language` as a fallback for `navigator.languages`, because of reports of ios 9 only supporting the singular variant (#1085).

I can't test this myself.


Maqsyo:
Thanks man!

friday:
Also thought about adding a fallback to `en`, but would like to know if this works first.

Maqsyo:
testet it and it is definetly the fix for this Problem!

PR #1092, opened 2018-07-09
Fix streaming link in docs

friday:
Change the streaming link to "Try Plyr online", since I removed the streaming section and forgot to change the link.

Made the base branch `master` since it's a doc change.

PR #1094, opened 2018-07-10
Verify internal documentation links with "remark"

friday:
Add markdown linting with remark and a plugin to check internal links.

It's easy to miss them when changing the docs (as demontrated by me in #1092). I also recall a couple other times though, and I found another when I had the linting in place (fixed in this PR)

Script can be run with `npm run remark` or `yarn remark`. It also runs with the normal `lint` script used by travis, but only when eslint doesn't fail. This can be changed so that both runs separately, but doing it in this PR would conflict with the `develop`-branch.
 
Could add more markdown linting. For example external links and general validation. This was just the "quick win" I was looking for.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
     },
     "scripts": {
         "build": "gulp build",
-        "lint": "eslint src/js",
+        "lint": "eslint src/js && npm run-script remark",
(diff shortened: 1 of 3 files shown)


sampotts:
Nice one! 👍

Issue #1119, opened 2018-07-24
YouTube's interface language param (hl) is not supported


renaudleo:
### Expected behaviour
When using the `hl` param in YouTube's iframe src, YouTube's player interface language should be set to the value of that param. See https://developers.google.com/youtube/player_parameters?playerVersion=HTML5#hl for reference.

### Actual behaviour
`hl` param is ignored. Passing the same param in a plain iframe embed without Plyr works fine.

### Environment
- Browser: Using Chrome, but isn't browser related.
- Version: 67.0
- Operating System: Mac OS
- Version: 10.13

### Steps to reproduce
- Add `hl=fr` to a YouTube embed URL
```html
<div class="plyr__video-embed" id="player">
    <iframe src="https://www.youtube.com/embed/bTqVqk7FSmY?hl=fr" allowfullscreen allowtransparency allow="autoplay"></iframe>
</div>
```

Looking at the code, this is most probably a quick fix. I think the only thing missing is to add the param with the other `playerVars` here: https://github.com/sampotts/plyr/blob/master/src/js/plugins/youtube.js#L189

PR on the way!

PR #1120, opened 2018-07-24
Add support for YouTube's hl param

renaudleo:
### Link to related issue
#1119 

Added support for YouTube's `hl` param, which allows setting the player's interface language.

While adding it, I noticed Plyr uses `url.searchParams.length` to check if there are any params in the URL. Since `searchParams` is an iterator, the `length` property doesn't exist, so this check always return false. I changed it to use `url.search` which returns a string version of all parameters (or an empty string).

#### Before
image

#### After (using `hl=fr`)
image

### Checklist

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -171,7 +171,7 @@ class Plyr {
                     this.elements.container.className = '';
 
                     // Get attributes from URL and set config
-                    if (url.searchParams.length) {
+                    if (url.search.length) {
                         const truthy = ['1', 'true'];
 
                         if (truthy.includes(url.searchParams.get('autoplay'))) {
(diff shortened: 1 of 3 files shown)


src/js/plyr.js
```diff
@@ -171,7 +171,7 @@ class Plyr {
                     this.elements.container.className = '';
 
                     // Get attributes from URL and set config
-                    if (url.searchParams.length) {
+                    if (url.search.length) {
```
sampotts: What's the reason for this change?

sampotts: Ah good spot. I knew about `.search` but never realised `.searchParams` had no `length` property.

Issue #1108, opened 2018-07-20
Youtube: Uncaught TypeError: Cannot redefine property: playbackRate


Scrity1976:
### Expected behaviour
No errors playing Youtube videos with Plyr

### Actual behaviour
Sometimes Youtube video plays without sound and throws error in console:
```
Uncaught TypeError: Cannot redefine property: playbackRate
    at Function.defineProperty (<anonymous>)
    at onReady (youtube.js:288)
    at M.h.G (www-widgetapi.js:48)
    at Y.h.o (www-widgetapi.js:92)
    at Y.h.H (www-widgetapi.js:105)
    at Wa.g (www-widgetapi.js:81)
    at Oa.f (www-widgetapi.js:70)
onReady @ youtube.js:288
h.G @ www-widgetapi.js:48
h.o @ www-widgetapi.js:92
h.H @ www-widgetapi.js:105
Wa.g @ www-widgetapi.js:81
Oa.f @ www-widgetapi.js:70
```

### Environment

- Browser: Chrome 67.0.3396.99 64 bit.
- Version: v.3.3.22
- Operating System: Windows 10 64 bit.

### Steps to reproduce
It may be difficult to reproduce this error.
- Embed Youtube video to Plyr.
- Reload page few times,

**Edit:** Also noticed, that not all Youtube videos have thumbnail images, but Plyr always tries to load them, and it causes `Cannot redefine property: playbackRate` error and `GET https://img.youtube.com/vi/2tOMS4OdO4c/maxresdefault.jpg 404 () Failed to load resource: the server responded with status of 404 (Not Found)`

**Edit 2:** Also noticed this error together with `Failed to execute 'postMessage' on 'DOMWindow': The target origin provided ('https://www.youtube.com') does not match the recipient window's origin ('http://...').
h.B @ www-widgetapi.js:99
`

friday:
I'm guessing onReady is being called twice for some reason. Pretty sure this is in no way connected to posters.

friday:
If someone getting this issue could help out debugging, that would be appreciated. I would try to find out why / if `onReady` is being called twice. There may be some issue with the setup method, or this is an actual YouTube issue.

Scrity1976:
Fixed it by adding check for `onReady` function, example:

```javascript
var isVideoReady = false;
onReady: function onReady(event) {
	if(!isVideoReady){
		isVideoReady = true;
		// onReady code here...
	}
}

friday:
[...]

PR #1131, opened 2018-07-28
Make sure youtube.onReady doesn't run twice

friday:

#1108

Adds a condition to youtube.onReady and bails if it has already been run (checking if it has `play`).

I haven't actually been able to replicate #1108, so I can't test that this works (only that it doesn't break). It should work, but in case the real issue (onReady being called twice) is also caused by our code, this is not an optimal fix.

### Checklist

### DIFF
--- a/src/js/plugins/youtube.js
+++ b/src/js/plugins/youtube.js
@@ -239,6 +239,10 @@ const youtube = {
                     triggerEvent.call(player, player.media, 'ratechange');
                 },
                 onReady(event) {
+                    // Bail if onReady has already been called. See issue #1108
+                    if (is.function(player.media.play)) {
+                        return;
+                    }
                     // Get the instance
                     const instance = event.target;
 


src/js/plugins/youtube.js
```diff
@@ -239,6 +239,10 @@ const youtube = {
                     triggerEvent.call(player, player.media, 'ratechange');
                 },
                 onReady(event) {
+                    // Bail if onReady has already been called. See issue #1108
+                    if (typeof player.media.play === 'function') {
```
sampotts: Should we use `is.function(player.media.play)`?

friday: Sorry! I forgot which repo I was in it seems ;)

friday: Fixed!

sampotts:
Cheers! 👍

PR #1134, opened 2018-07-29
Hide empty controls

mjfwebb:
None.

If the plyr__controls is empty it is still showing the fading transition causing captions to be pushed up when hovering over where the controls would be. This change hides the plyr__controls div when it is empty.

### Checklist

### DIFF
--- a/src/js/controls.js
+++ b/src/js/controls.js
@@ -1455,12 +1455,9 @@ const controls = {
             target = this.elements.container;
         }
 
-        // Inject controls HTML
-        if (is.element(container)) {
-            target.appendChild(container);
-        } else if (container) {
-            target.insertAdjacentHTML('beforeend', container);
-        }
+        // Inject controls HTML (needs to be before captions, hence "afterbegin")
+        const insertMethod = is.element(container) ? 'insertAdjacentElement' : 'insertAdjacentHTML';
+        target[insertMethod]('afterbegin', container);
 
         // Find the elements if need be
         if (!is.element(this.elements.controls)) {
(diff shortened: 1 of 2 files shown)


friday:
Travis unfortunately doesn't understand when you change the base branch, so the script still fails although it's fixed https://travis-ci.org/sampotts/plyr/jobs/409611204#L460

mjfwebb:
You're right. The captions are being moved still. I'll take another look.

friday:
Pushed a commit that changes the lower controls to be injected before the captions + expanded on your existing change to override the positioning.

Should do the trick, and since the lower controls are positioned absolutely it doesn't seem to create any issues.

mjfwebb:
I think it needs to be set to `transform: translateY(-15px);` instead of `transform: translateY(0px);`.

sampotts:
Any non zero values should come from a variable really if we're changing that. I'll look at tonight and merge.

mjfwebb:
image

There's a 15 pixel difference at the moment. In the interest of getting it 'perfect', the CSS change would be to increase the padding to 35px.

```
.plyr__controls:empty ~ .plyr__captions {
	padding: 35px;
}
```

Example:

https://codepen.io/mjfwebb/pen/GBOLJQ

friday:
Why do you want the captions to float 35px above the bottom when there are no controls?

mjfwebb:
Because that's the position they are in when controls are hidden currently. I don't mind them being in a different position at all, I just wanted it to be uniform.

PR #1138, opened 2018-07-30
Size plyr container with width="size" value of Video element

lookingdown:
Makes plyr containers same width as originating <video> element "width" attribute works for any number of videos on same page fullscreen escape click and esc button also works, I needed this to have plyr working as <video> element in Tinymce and be treated as such and be able to set  size of video in <video> element attribute something which would be logical for starters :)..

### Checklist
- [ ] Use `develop` as the base branch

### DIFF
--- a/src/js/fullscreen.js
+++ b/src/js/fullscreen.js
@@ -50,10 +50,17 @@ function toggleFallback(toggle = false) {
 }
 
 class Fullscreen {
+    var fWidth = '';
     constructor(player) {
         // Keep reference to parent
         this.player = player;
-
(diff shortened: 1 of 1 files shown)


friday:
Thanks!

However I'm not sure the behaviour is something we want in Plyr. Even if we do, I can't accept this code, since it breaks our guidelines, it doesn't lint, and seems very bloated. It's also made to the wrong branch. Also, we don't have any `_this`, so I don't think it can work?

Please create an issue explaining the problem clearly and we can take the discussion from there.

https://github.com/sampotts/plyr/blob/master/contributing.md#contributing-features-and-documentation

Issue #1137, opened 2018-07-30
Captions position mismatch on various control schemes


mjfwebb:
### Expected behaviour

Captions remain in the same position when no controls are shown (they could be auto-hidden), regardless of controls being configured or not.

### Actual behaviour
Captions are shown in different positions (y-axis) when controls are hidden versus when no controls are configured.

### Steps to reproduce
Configure a player with only `controls: ['play-large']` or `controls: []`

### Link to where the bug is happening

https://codepen.io/mjfwebb/pen/PBEaRY
(Should autoplay if you have vimeo muted. If not, load the codepen then refresh after a couple of seconds.)

friday:
Thanks! This actually shows three separate problems.

1. The first and second video position captions differently.
2. The third example shows a similar bug with `null` added to the output that was fixed in #868.
3. The third example has no controls element at all. The css we added is applied for captions *after* empty controls, but not after "no controls". So the css should be changed.

#1139 should fix 2-3 (not verified with your example though).

This can be used to fix 1:

```css
.plyr--hide-controls .plyr__captions {
 transform: translateY(0);
}
```

friday:
This is fixed now (but not released yet). I made a temporary build to test with your pen. If you want, you can verify it by changing the external references in your pen's settings to:
https://rawgit.com/friday/plyr/1137-build/dist/plyr.min.js
https://rawgit.com/friday/plyr/1137-build/dist/plyr.css

mjfwebb:
Agreed, I updated the pen and it's working excellently now. Thanks @friday

PR #1139, opened 2018-07-30
Controls input fixes

friday:
Fixes some issues with the `controls` option.
* It should support a function that returns an array ([this is documented](https://github.com/sampotts/plyr/blob/master/controls.md#controls)).
* The documentation should include the `Element` type.
* `null` shouldn't be printed when the controls are `[]` (regression in my e5e169a, see [test pen](https://codepen.io/fullkornslimpa/pen/NBXOXG)).
* There should always be an element for the lower controls, even if there aren't any lower controls (fixes the positioning for third example in the #1137 codepen). Note that there are other ways to fix this issue, for example changing the css. The reason I went with this is that it also fixed the previous issue ^ and I think `controls.create()` shouldn't output null.

In addition
* If `controls` is falsy, the menu will now be empty (`undefined` shouldn't be falsy due to how the options are extended unless `Plyr.defaults.controls` is also unset). 

### Checklist
- [x] Test on Chrome only since the changes are not browser-specific or prone to browser issues.

### DIFF
--- a/controls.md
+++ b/controls.md
@@ -3,7 +3,9 @@
 This is the markup that is rendered for the Plyr controls. You can use the default controls or provide a customized version of markup based on your needs. You can pass the following to the `controls` option:
 
 -   `Array` of options (this builds the default controls based on your choices)
+-   `Element` with the controls
 -   `String` containing the desired HTML
+-   `false` (or empty string or array) to disable all controls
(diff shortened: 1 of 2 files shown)


sampotts:
Cheers! 👍

PR #1141, opened 2018-07-31
Improve captions positioning consistency

friday:
#1137

Makes sure the captions are only repositioned when there is a non-empty, non-hidden lower menu (rather than making css exceptions later).

Also moved up the rule to hide the empty controls and added a comment, as I think this is an important rule to the semantics.

### Checklist
- [x] Test on Crome only and checked caniuse that these changes should be safe for IE9+
- [ ] Actually tested the above


sampotts:
Cheers 👍

PR #1158, opened 2018-08-12
Fix #1153: Captions language fallback

### LINKED ISSUE #1153: bug in caption.language initialization
### Expected behaviour
`player.captions.language` should be `en`

### Actual behaviour
`player.captions.language` is `e`

### Steps to reproduce
update the browser setting, make sure `navigator.languages || navigator.language || navigator.userLanguage` return a string instead of an array, which is what i got from IE10 in Win7

### Environment

- Browser:
Internet Explorer
- Version:
11
- Operating System:
Windows
- Version:
7

there is a problem in this code: https://github.com/sampotts/plyr/blob/master/src/js/captions.js#L87, it assume `navigator.languages || navigator.language || navigator.userLanguage` always return an array of strings but it is not, when `navigator.languages` is empty it returns a string then with `Array.from` it becomes array of characters.
It could be `Array.from([...(navigator.languages || []), navigator.language, navigator.userLanguage]).map(language => language.split('-')[0])` to avoid this issue.

friday:

#1153

Changes the logic to fix the issue.

### Checklist

### DIFF
--- a/src/js/captions.js
+++ b/src/js/captions.js
@@ -83,11 +83,8 @@ const captions = {
         // * active:    The state preferred by user settings or config
         // * toggled:   The real captions state
 
-        const languages = dedupe(
-            Array.from(navigator.languages || navigator.language || navigator.userLanguage).map(
-                language => language.split('-')[0],
-            ),
-        );
+        const browserLanguages = navigator.languages || [navigator.language || navigator.userLanguage || 'en'];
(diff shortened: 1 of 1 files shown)


sampotts:
Thanks! 👍 I'll do a release later today with the a11y stuff since there's minor stuff remaining.

PR #1175, opened 2018-08-24
change getAttribute "size" to "data-res" on html5.js

halionsp:
- Attribute size not allowed on element source at this point
→https://github.com/sampotts/plyr/issues/1172←

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -94,9 +94,9 @@
             <div id="container">
                 <video controls crossorigin playsinline poster="https://cdn.plyr.io/static/demo/View_From_A_Blue_Moon_Trailer-HD.jpg" id="player">
                     <!-- Video files -->
-                    <source src="https://cdn.plyr.io/static/demo/View_From_A_Blue_Moon_Trailer-576p.mp4" type="video/mp4" data-res="576">
-                    <source src="https://cdn.plyr.io/static/demo/View_From_A_Blue_Moon_Trailer-720p.mp4" type="video/mp4" data-res="720">
-                    <source src="https://cdn.plyr.io/static/demo/View_From_A_Blue_Moon_Trailer-1080p.mp4" type="video/mp4" data-res="1080">
+                    <source src="https://cdn.plyr.io/static/demo/View_From_A_Blue_Moon_Trailer-576p.mp4" type="video/mp4" data-size="576">
+                    <source src="https://cdn.plyr.io/static/demo/View_From_A_Blue_Moon_Trailer-720p.mp4" type="video/mp4" data-size="720">
+                    <source src="https://cdn.plyr.io/static/demo/View_From_A_Blue_Moon_Trailer-1080p.mp4" type="video/mp4" data-size="1080">
 
                     <!-- Caption files -->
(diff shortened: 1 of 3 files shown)


sampotts:
Why not data-size? Also, this way it’s a breaking change as it’s not backwards compatible. I can’t merge this as is unless it waits for v4.0.0

halionsp:
Sorry not merge. The commit https://github.com/sampotts/plyr/commit/a1fea91794ce527ae08d656c7dab7e1cfc28ecc5  does not work. 💀

friday:
> Why not data-size?

Sorry. I encouraged `data-res` in the issue. `size` is ambiguous imo (could just as well mean file size or bit rate).

halionsp:
> Sorry. I encouraged data-res in the issue. size is ambiguous imo (could just as well mean file size or bit rate).

It is right. I think the same.  The name of `data-size` is better. And it would be available for the major version 4.

```
diff --git a/src/js/html5.js b/src/js/html5.js
index fc8da8c..af3cb20 100644
--- a/src/js/html5.js
+++ b/src/js/html5.js
@@ -23,7 +23,7 @@ const html5 = {
         // Get sizes from <source> elements
         return html5.getSources
             .call(this)
-            .map(source => Number(source.getAttribute('size')))
+            .map(source => Number(source.getAttribute('data-size')))
             .filter(Boolean);
     },
 
@@ -42,14 +42,14 @@ const html5 = {
                 const source = sources.find(source => source.getAttribute('src') === player.source);
 
                 // Return size, if match is found
-                return source && Number(source.getAttribute('size'));
+                return source && Number(source.getAttribute('data-size'));
             },
             set(input) {
                 // Get sources
                 const sources = html5.getSources.call(player);
 
                 // Get first match for requested size
-                const source = sources.find(source => Number(source.getAttribute('size')) === input);
+                const source = sources.find(source => Number(source.getAttribute('data-size')) === input);
 
                 // No matching source found
                 if (!source) {

```

arankaren:
> Sorry. I encouraged data-res in the issue. 

I also think that `data-res` is a good idea for the major version 4

```
diff --git a/src/js/html5.js b/src/js/html5.js
index fc8da8c..af3cb20 100644
--- a/src/js/html5.js
+++ b/src/js/html5.js
@@ -23,7 +23,7 @@ const html5 = {
         // Get sizes from <source> elements
         return html5.getSources
             .call(this)
-            .map(source => Number(source.getAttribute('size')))
+            .map(source => Number(source.getAttribute('data-res')))
             .filter(Boolean);
     },
 
@@ -42,14 +42,14 @@ const html5 = {
                 const source = sources.find(source => source.getAttribute('src') === player.source);
 
                 // Return size, if match is found
-                return source && Number(source.getAttribute('size'));
+                return source && Number(source.getAttribute('data-res'));
             },
             set(input) {
                 // Get sources
                 const sources = html5.getSources.call(player);
 
                 // Get first match for requested size
-                const source = sources.find(source => Number(source.getAttribute('size')) === input);
+                const source = sources.find(source => Number(source.getAttribute('data-res')) === input);
 
                 // No matching source found
                 if (!source) {

```

PR #1196, opened 2018-09-29
Fix keyboard focus

TechGuard:
Focused keyboard input did not register when clicking on the video. This makes the container focusable so that it will focus the keyboard input wherever you click on the video.

### DIFF
--- a/src/js/listeners.js
+++ b/src/js/listeners.js
@@ -266,6 +266,7 @@ class Listeners {
         // Keyboard shortcuts
         if (!player.config.keyboard.global && player.config.keyboard.focused) {
             on.call(player, elements.container, 'keydown keyup', this.handleKey, false);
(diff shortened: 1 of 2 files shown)


sampotts:
Pull requests should be to the develop branch as per docs

sampotts:
If you click a focusable element, it automatically gets focus, so much of that code is redundant. Also, as part of the acccesibility issue, tabindex was removed from the container. What’s the actual bug here?

TechGuard:
Okay I didn't know you couldn't use tabindex for the container. But here's the issue that I had.

When I click on the video to pause it or play the video. It does **not** have keyboard focus and my keyboard shortcuts do not work.
My change also sets the focus on the player whenever you have autoplay enabled. So that you can use the shortcuts right away.

sampotts:
The `tabindex` has been restored in v3.5.0 which should resolve this 👍

PR #1197, opened 2018-09-29
Fix html5 quality settings

TechGuard:
Only save quality setting when it's updated by the user. Fixes bug in html5 player where it would override the settings if the current video does not support the given quality.


sampotts:
Again, I’m not seeing the issue here. You’ll notice the logic is that if the quality level passed is not supported, we choose the closest supported level. Do you have a code pen demonstrating the issue with the current implementation?

TechGuard:
I don't know what went wrong when I branched the project. It has all the changes from 'develop'.

Here's the problem that I have with the quality settings.

1. Load a video that supports 1080p and update the settings
2. Load a video that does **not** support 1080p and it will override your settings to the next available quality ie. 720p
3. Go back to the last video that supports 1080p and you now watch the video in 720p

My pull request will only save the quality settings whenever the user changes them. Not the player itself.

sampotts:
Thanks for this 👍

PR #1211, opened 2018-10-13
duration after changing video quality

melbahja:
continue at the current duration after changing video quality

### Checklist
- [x] Test on: chrome, firefox

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -680,8 +680,7 @@ class Plyr {
     set quality(input) {
         const config = this.config.quality;
         const options = this.options.quality;
-        const duration = this.duration;
-        const isPlaying = this.playing;
+        const { duration, playing } = this;
 
         if (!options.length) {
             return;
@@ -706,11 +705,11 @@ class Plyr {
         // Set quality
         this.media.quality = quality;
         
-        // seek to duration before changing quality
+        // Seek to duration before changing quality
         this.seek = duration;
 
-        // continue
-        if (isPlaying) {
+        // Continue
+        if (playing) {
             this.play();
         }
     }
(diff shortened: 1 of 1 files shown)


src/js/plyr.js
```diff
@@ -680,6 +680,8 @@ class Plyr {
     set quality(input) {
         const config = this.config.quality;
         const options = this.options.quality;
+        const duration = this.duration;
+        const isPlaying = this.playing;
```
sampotts: Can we not use object destructuring here? e.g. 

`const { duration, playing } = this;`

src/js/plyr.js
```diff
@@ -703,6 +705,14 @@ class Plyr {
 
         // Set quality
         this.media.quality = quality;
+        
+        // seek to duration before changing quality
```
sampotts: It'd be good to start comments with a capital to match the others

sampotts:
Thanks for this 👍

sampotts:
Remind me why we want to seek to the end of the video on source change? I've backed this one out as I don't think there's any benefit to it. What was the issue you were trying to solve?

melbahja:
@sampotts when you change the video quality plyr starts from the beginning
This PR make the video continue from the point before changing the quality

sampotts:
I can't reproduce that. Anyway `duration` is the _length_ of the media, I think you meant `currentTime` but that's already handled. For me it seeks back to where it was previously.

melbahja:
yes its `currentTime` before v3.4.5 there was an issue when change the quality. now i saw plyr.io demo it's working fine

sampotts:
OK. Yeah, your code was using `duration` instead of `currentTime` ...

```javascript
// Seek to duration before changing quality
this.seek = duration;
```

Anyway, glad it's sorted. Nothing has changed with that recently. Perhaps you were using an old version.

PR #1226, opened 2018-10-22
Prevent immediate hiding of controls on mobile

jamesoflol:
NA. Chatted about it on slack

Current behaviour: When a touch device user taps/slides the seek bar, the controls disappear immediately. This doesn't happen on non-touch, thanks to controls.hover.

Issue: While this is ideal on play event, it's not ideal for seek:
1. UX: While seeking, a user doesn't know exactly where they want to land, and often has to seek multiple times. It's a bit jarring to have the seek bar disappear immediately.
2. If the player needs to load after seeking, the current functionality causes the controls to disappear, , reappear while loading, then disappear again immediately after. When all this happens in the space of a second, it just looks like a flicker/bug.

Solution: Prevent hiding of controls within 2 seconds of seeking.
I've implemented this is a timer. This doesn't feel ideal, but it was much messier to do it otherwise. This is because checkPlaying() and checkLoading() are fired 4-6 times after seeking, and we don't want to just prevent toggleControls() from these functions entirely on mobile, as that would prevent hiding controls on play(). The timer of 2 seconds was chosen because it's right in-between the immediate show attempts, and the existing 4-second hide timer.

### Checklist

### DIFF
--- a/src/js/listeners.js
+++ b/src/js/listeners.js
@@ -620,6 +620,9 @@ class Listeners {
                 return;
             }
 
+            // Record seek time so we can prevent hiding controls for a few seconds after seek
+            player.lastSeekTime = Date.now();
+
             // Was playing before?
             const play = seek.hasAttribute(attribute);
 
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -302,6 +302,9 @@ class Plyr {
         if (this.config.autoplay) {
             this.play();
         }
+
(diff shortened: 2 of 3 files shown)


sampotts:
This looks great! I'll merge it later today and get a new version out along with the download button once I remember where I was up to with that.

trafium:
If I understand correctly, controls are still being hidden on forward/rewind button taps.

jamesoflol:
Right you are. I didnt even know about forward/back buttons. Should be logical and a small matter to update lastSeekTime during those events too?

Issue #1159, opened 2018-08-12
Seek-Preview


calvinc64:
I want to add something like YouTube has for quite some time now, I want you to add a Seek-Preview for me like on YouTube, because I am recently quite interested in dtube project but the player pisses me so... here is my thing to ask if it is possible:

We have this Preview Image, Prerendered: https://ipfs.io/ipfs/QmaJNsdbT2jARQYSmNJuNhCQucLiTmWGookzCFjjSpzpLq

and this is the thing I want: 
![image](https://user-images.githubusercontent.com/5560381/44006187-034a4924-9e80-11e8-995a-1475c9629c27.png)

thanks in advance.
Sorry for my English.

Ulf3000:
1up for this , without the preview , every alternative youtube player is automatically worthless , sorry but thats the truth

sampotts:
I’ll still sleep at night don’t worry. Feel free to get a pull request in...

Offbeatmammal:
looks like it would be possible to add something into `controls.js` in `updateSeekTooltip(event)` to also look up a thumbnail image for the relevant timestamp and show it while scrubbing on the seek bar but it's not as trivial/easy as just 'add some hover-over iamges':
- can we query those images from youTube?
- can we query those images from Vimeo?
- in all cases what granularity of thumbnails do we need (eg is it one every second, or one every _n_% of the length of the content)
- how would they be generated/named (ffmpeg for the first, number the naming convention is going to tie back to how they're used/referenced)?

jamesoflol:
There's already a PR: https://github.com/sampotts/plyr/pull/1253

I was hoping for a little testing/feedback from anyone before putting the finishing touches on it

In answer to your questions:
- Thumb file naming: I've gone with the VTT spec for thumbs, same as a few other implementations out there, so that people can use any interval + naming convention that they want
- Granularity: One every n seconds. Different intervals depending on video length. My personal preference is:
  - > 1hr: 10 second interval
  - > 20m: 5 second interval
  - > 5m: 2 second interval
  - < 5m: 1 second interval

Offbeatmammal:
@jamesoflol - quick look at the implementation and it's really nice! using vtt makes it more flexible (I've done an implementation in the past that embed the tc into the filename which was messy!)
quick play on macOS browsers (current Chrome, Firefox, Safari) I didn't see any issues.
Agree sprite images would be nice (though initial load might be slower, so would need to preload/pre-cache the image.
[...]

Issue #46, opened 2015-03-03
WebVTT files to load Tooltip Thumbnails


adantzler:
Plyr.io looks awesome!  Great work!
When we encode our videos we generate a VTT file and include thumbnails in it.

We are currently using JWPlayer and it's format we follow is below:

WEBVTT

00:00.000 --> 00:05.000
/assets/preview1.jpg

00:05.000 --> 00:10.000
/assets/preview2.jpg

00:10.000 --> 00:15.000
/assets/preview3.jpg

00:15.000 --> 00:20.000
/assets/preview4.jpg

SamPotts:
Hey @adantzler,

Thanks! Nice idea. I'll have a look at implementing something for sure. :+1:

adantzler:
That's awesome, let me know if I can help in any way.

hookro:
I've created a little snippet to catch displayed seek onHover :dancer: 
```
var video = document.getElementById("thumb");
video.addEventListener("loadedmetadata", getSeeker);

		function getSeeker(){
            
            var seekTime;
            
            $('.plyr__progress--seek').mousemove(function() {
                if (enableHandler) {
                    $(".plyr__tooltip.plyr__tooltip--visible").on('DOMSubtreeModified', function(){
                       seekTime = ($(this).html());
                        // add Thumbnail DIV at the requested time
                    });
                    enableHandler = false;
                }
            });
            
            $('.plyr__progress--seek').mouseout(function() {
				// >>remove thumbnail div correctly
            });
			
		  timer = window.setInterval(function(){
                      enableHandler = true;               
            }, 100);
        }
```

SamPotts:
Cool! You might want to use MutationObserver rather than DOM events given the performance benefit:
https://developer.mozilla.org/en/docs/Web/API/MutationObserver

hookro:
Thanks! I didn't looked at performance, because I did not finished yet. I will take a look at MutationObserver, when will be ready and works well for me, I will post it here. 
I think it will be easy if you could provide seekTime either than grab it, but It's not a hurry.

Here is a dynamic thumbnail, I don't think its powerful, I would recommend server-side thumbnails.


HTML code
```
<canvas id="canvas" 
        width="750px" height="540px"
        style="display:block;">
</canvas>
<div id="screenShots"></div>
```
JS code
```
        var video = document.getElementById("thumb");
		video.addEventListener("loadedmetadata", initScreenshot)
		video.addEventListener("playing", startScreenshot);
[...]

PR #1253, opened 2018-11-12
Preview seek/scrubbing thumbnails

jamesoflol:
https://github.com/sampotts/plyr/issues/46
https://github.com/sampotts/plyr/pull/376
https://github.com/sampotts/plyr/issues/1159

When user hovers seek bar, they see a preview image
When they drag the seek bar (scrubbing) they see a larger preview image

### Checklist
- [x ] Use `develop` as the base branch
- [x ] Exclude the gulp build (`/dist` changes) from the PR
- [x ] Test on [supported browsers](https://github.com/sampotts/plyr#browser-support)

### Demo:
https://s3-ap-southeast-2.amazonaws.com/jameslol-test/plyr-preview-thumbs/demo/index.html

### Notes:

 * Settings:
    previewThumbnails: {
        enabled: true,
        src: [
          'https://dev.videohop.com.au/s3/v/bluemoon/100p/100p.vtt',
          'https://dev.videohop.com.au/s3/v/bluemoon/240p/240p.vtt',
        ]
    }
 * Allows multiple qualities
   - Each VTT file MUST share the same interval/timings. it would cause silliness to have different timings. and the current logic for progressing up through qualities relies on each quality having the same index
   - While hovering the same time, both small thumb and fullsize will progressively load higher qualities in order until the image's naturalHeight is greater than or equal to the contain height. (or of course, until we hit the largest that we have)
 * I'd recommend increasing the height of the seek bar - it's currently a bit thin for easily keeping the preview showing. Something like
.plyr__progress input[type=range] {
    padding-top: 15px;
    padding-bottom: 15px;
}
 * Thumbs are set via JS settings on Plyr init, not HTML5 'track' property. Using the track property would be a bit gross, because it doesn't support custom 'kinds'. kind=metadata might be used for something else, and we want to allow multiple thumbnails tracks. Tracks must have a unique combination of 'kind' and 'label'. We would have to do something like kind=metadata,label=thumbnails1 / kind=metadata,label=thumbnails2. Square peg, round hole
 * It currently works fantastically with jpeg sprites. If you use individual images instead, ymmv. There's no perfect solution to using individual images, and it has many downsides.

### Todo:
- Add to readme
  - Include notes/specs on jpeg sprites

### Todo later:
- See if the vtt parser function accepts comments https://developer.mozilla.org/en-US/docs/Web/API/WebVTT_API
  - Also, hours is optional? (can be omitted instead of writing '00:')

### DIFF
--- a/src/js/config/defaults.js
+++ b/src/js/config/defaults.js
@@ -342,6 +342,8 @@ const defaults = {
         loading: 'plyr--loading',
         hover: 'plyr--hover',
         tooltip: 'plyr__tooltip',
+        previewThumbnailContainer: 'plyr__preview-thumbnail-container',
+        previewScrubbingContainer: 'plyr__preview-scrubbing-container',
         cues: 'plyr__cues',
         hidden: 'plyr__sr-only',
         hideControls: 'plyr--hide-controls',
@@ -395,6 +397,11 @@ const defaults = {
         enabled: false,
         publisherId: '',
     },
+
+    // Preview Thumbnails plugin
+    previewThumbnails: {
+        enabled: false,
+    }
 };
 
 export default defaults;
--- a/src/js/config/defaults.js
+++ b/src/js/config/defaults.js
@@ -342,8 +342,6 @@ const defaults = {
         loading: 'plyr--loading',
         hover: 'plyr--hover',
         tooltip: 'plyr__tooltip',
-        previewThumbnailContainer: 'plyr__preview-thumbnail-container',
-        previewScrubbingContainer: 'plyr__preview-scrubbing-container',
         cues: 'plyr__cues',
         hidden: 'plyr__sr-only',
         hideControls: 'plyr--hide-controls',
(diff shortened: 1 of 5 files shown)


jamesoflol:
Have updated the code to use jpeg sprites, and have generated jpeg sprites for the blue moon demo video, and updated my demo page accordingly. It is much snappier now.

jamesoflol:
- Fixed bug: Edge seek errors: Replaced array spread with Array.from()
- Fixed IE11 bug: seek time was offset to the left. Required an extra container div to facilitate this

sampotts:
Thanks! I'll give this a go locally

sampotts:
Looks great! I've shot you a message on Slack about getting the images you've created and the VTT so I can add them to the demo. I'll use some of those notes you've written for the docs.

regiside:
This is such a great feature! Thank you so much James for this top notch implementation, it really takes Plyr to the next level. I just finished implementing a feature that automatically generates the sprite sheets and VTTs from vimeo videos on our site using a CLI tool called Movie Thumbnailer (mtn) and PHP, and I'd like to make a video explaining how to put it all together. I'm wondering though if there's any way to adjust the size that the thumbnails get rendered at. Right now it seems like the box is about 200px wide and I'd like something a little smaller like 160px. Thanks!  

-Reggie

Issue #858, opened 2018-04-03
Clicking on video does not trigger custom listeners


gurupras:
### Expected behaviour
Clicking video toggles play/pause as expected, but does not trigger functions registered via `listeners`

### Actual behaviour
Expect functions registered via `listeners` to trigger

### Environment
- Browser: Chrome/Firefox
- Version:
  - Chrome: 65.0.3325.181
  - Firefox: 59.0.1
- Operating System: Windows 10
- Version: Build 1703

### Steps to reproduce
 - Create plyr instance with a custom `play` listener
 - Click on the video to toggle play/pause
 - Observe that the custom handler you registered is not being called
Here's a [pen](https://codepen.io/gurupras/pen/OvEwyw) illustrating the problem

gurupras:
I missed this while out of town. This issue is still present as of v3.3.21. Here's a [pen](https://codepen.io/gurupras/pen/BVbQjZ)

A slightly related issue that I will point out (while attempting to fix this) is that although the documentation lists `pause` as a valid event that we can bind an early-listener to, it seems impossible for me to get the code to trigger this in the current implementation.

friday:
[Reduced test case](https://css-tricks.com/reduced-test-cases/):
https://codepen.io/anon/pen/ERJKmM?editors=0011

SayGoodBey:
hi,I have a problem now. Listen for the return result of asynchronous function to judge return. For example, interact with native to get network status to decide whether to play or not. How to deal with this block?

CyanoFresh:
also clicking on play button when video is playing triggers play event, instead of pause event

gurupras:
Ah yes, this is because there is no `pause` event. Not sure if this change was documented (or intended).

Regardless, you handle it like this now:  

```
plyr = new Plyr(el, {
  listeners: {
    play (e) {
      if (self.plyr.playing) {
        ...
      } else {
        ...
      }
    }
  }
})

sampotts:
Do you mean _event_ or _listener_ ?

gurupras:
Sorry, you're right. I meant to say that there is no _listener_ for the `pause` event anymore. The `play` listener can however be multiplexed to handle pauses as shown above.

CyanoFresh:
yep, need to use workaround for pause listener. maybe remove it from defaults.js? it's confusing now

sampotts:
My preference would be to, you know, fix it 👍

sampotts:
[...]

PR #1267, opened 2018-11-20
Ensure custom handlers are called on container clicks that trigger togglePlay or restart

gurupras:
This PR fixes issue #858 --- Custom listeners were not being called when the plyr container was clicked (triggering play/pause).

  - Replaced calls to `player.restart()` and `player.togglePlay()` with `proxy(...)`

### Checklist

I don't know how to test on _all_ the supported browsers. So far, I have tested it on Chrome and Firefox and it works.

### DIFF
--- a/src/js/listeners.js
+++ b/src/js/listeners.js
@@ -387,7 +387,7 @@ class Listeners {
 
                 if (player.ended) {
                     this.proxy(event, player.restart, 'restart');
-                    this.proxy(event, player.togglePlay, 'play');
+                    this.proxy(event, player.play, 'play');
                 } else {
                     this.proxy(event, player.togglePlay, 'play');
                 }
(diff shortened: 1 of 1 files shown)


src/js/listeners.js
```diff
@@ -386,10 +386,10 @@ class Listeners {
                 }
 
                 if (player.ended) {
-                    player.restart();
-                    player.play();
+                    this.proxy(event, player.restart, 'restart');
+                    this.proxy(event, player.togglePlay, 'play');
```
sampotts: Shouldn't this one be `player.play`?

gurupras: You're right. Fixed and pushed

sampotts:
Changed base to develop (as per all PRs other than docs) and merged 👍

sampotts:
Changed base to develop (as per all PRs other than docs) and merged 👍

PR #1269, opened 2018-11-21
Add support for central translation repository

### LINKED ISSUE #1261: Central translation repository
Currently, the only possible way to use translation for Plyr is to get `i18n` object from somewhere and add it to Plyr. This have several weaknesses because users (developers) would need to know location of file for each translation.

There should be some central translation "repository" (the best options is some folder in this repository, for example `src/locales`) which would have all translations in JSON files.
Users (developers) would than just need to specify language name to get translation.

filips123:
It adds support for central translation repository and fixes #1261.

Translations are saved into `src/locales/{lang}.js` (where `{lang}` is ISO 639-1 code) and bundled to Plyr file.
Language could be specified with `language` options and it could not be changes by user (for now).
Translation could still be overwritten by `i18n` option.

In the future, you should add support for user changing of language in UI options. I've already added language name field for this (in `src/locales/index.js`).

You should also find translators to add translations. I added translation guide to `contributing.md` and I could translate UI to Slovenian.


src/js/utils/i18n.js
```diff
@@ -21,6 +22,23 @@ const i18n = {
             return '';
         }
 
+        const english = languages.filter(function (elem) {
```
sampotts: Can you make sure you have prettier and ESLint setup? This could be a one liner. Also you should use `===` rather than `==`. Basically this line can be:

```javascript
const english = languages.filter(e => e.code === 'en');
```

Also should we not use a config option for the default? It may not always be 'en'.

src/js/utils/i18n.js
```diff
@@ -21,6 +22,23 @@ const i18n = {
...
+            return elem.code == 'en';
+            }
+        );
+
+        const locale = languages.filter(function (elem) {
```
sampotts: As above. There's some odd syntax and formatting going on.

src/locales/en.js
```diff
@@ -0,0 +1,51 @@
+// ==========================================================================
+// Plyr English translation
+// ==========================================================================
+
+const i18n = {
+  restart: 'Restart',
```
sampotts: You'll notice the rest of the project uses 4 space tabs. We should stick to this 👍

src/locales/sl.js
```diff
@@ -0,0 +1,51 @@
+// ==========================================================================
+// Plyr Slovenian translation
+// ==========================================================================
+
+const i18n = {
+  restart: 'Ponovno predvajaj',
```
sampotts: As above with formatting

contributing.md
```diff
@@ -38,3 +39,15 @@ Please follow the instructions in our issue templates. Don't use github issues t
 * When finished, push the changes to your GitHub repository and send a pull request to **develop**. Describe what your PR does.
 
 * If the Travis build fails, or if you get a code review with change requests, you can fix these by pushing new or rebased commits to the branch.
+
+## Contributing translations
+
+* Fork Plyr, and create a new branch in your fork, based on the **develop** branch
+
+* Make and translate language file `src/locales/{lang}.js`, where `{lang}` is two-letter language code from [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes). It is recommended to copy and edit English translation.
```
sampotts: Do we not want to allow `en-AU`, `en-US`, `en-GB` etc?

sampotts:
Perhaps we just create a build task that accepts an argument for an ISO code and bundles that lang (if it exists) with the rest of the JS.

src/js/utils/i18n.js
filips123: Default language should be `en` because it is only for times, when translation doesn't have some translated strings, so like fallback language.

But maybe there could be support to change fallback language.

contributing.md
filips123: Which standard should be used for this? Code potentially allows such languages so only standard should be changed.

filips123:
> Perhaps we just create a build task that accepts an argument for an ISO code and bundles that lang (if it exists) with the rest of the JS.

Unfortunately I can't create custom build tasks because I (currently) don't know much about Node.js build systems and Gulp.

src/js/utils/i18n.js
```diff
@@ -21,6 +22,15 @@ const i18n = {
             return '';
         }
 
+        const english = languages.filter(e => e.code === 'en');
```
filips123: From https://github.com/sampotts/plyr/pull/1269#pullrequestreview-182933266:

> Also should we not use a config option for the default? It may not always be 'en'.

filips123: Default language should be `en` because it is only for times, when translation doesn't have some translated strings, so like fallback language.

But maybe there could be support to change fallback language.

src/js/utils/i18n.js
filips123: One liner is fixed.

Discussion for default language is moved to https://github.com/sampotts/plyr/pull/1269#pullrequestreview-187878861.

contributing.md
HybridSolutions: you should use that because for isntance pt-PT is different from pt-BR.

filips123: @HybridSolutions @sampotts Will [IETF language tag](https://en.wikipedia.org/wiki/IETF_language_tag) be OK for this?

HybridSolutions: I think so.

filips123: @HybridSolutions I will update this until end of this week.

filips123: @sampotts @HybridSolutions I added IETF language tag details.

filips123:
@sampotts Ping.

marcelogarbin:
any news? thank's

PR #1295, opened 2018-12-20
fix: Use Math.trunc instead of parseInt

taion:
`parseInt` coerces its input to a string before attempting to parse it. This means that, for very small fractions that stringify as something like "5e-7", these `parseInt` calls will not actually truncate the values to 0 as presumably desired, but will give the leading digit.

This leads to our seeing the formatted current time shoot up to a large value immediately after the video loops, as the next indicated time becomes a small enough number that, after being converted from seconds to hours, is small enough to trigger that issue with scientific notation above.

This replaces the usages of `parseInt` for time formatting with `Math.trunc`, which has the right behavior. There's another use for parsing the progress in the Vimeo plugin where I'm not too sure what's going on. It looks like that usage can also be safely replaced with either `Math.trunc` or `Math.floor`, but I'm not sure if I'm missing something there; it's not clear to me why that progress needs to be rounded or parsed in the first place.

Also, before doing browser testing, I wanted to confirm that it's okay to use an ES6 feature that isn't supported in the full browser matrix without polyfills.

### Checklist

### DIFF
--- a/src/js/utils/time.js
+++ b/src/js/utils/time.js
@@ -5,9 +5,9 @@
 import is from './is';
 
 // Time helpers
-export const getHours = value => parseInt((value / 60 / 60) % 60, 10);
-export const getMinutes = value => parseInt((value / 60) % 60, 10);
-export const getSeconds = value => parseInt(value % 60, 10);
+export const getHours = value => Math.trunc((value / 60 / 60) % 60, 10);
+export const getMinutes = value => Math.trunc((value / 60) % 60, 10);
+export const getSeconds = value => Math.trunc(value % 60, 10);
 
 // Format time to UI friendly string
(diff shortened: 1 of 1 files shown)


sampotts:
Thanks! I never knew about this. Very interesting. I guess it's recommended to use `Math.trunc` instead of `parseInt` generally? 

With regards to Vimeo progress, we could probably do a string comparison I guess. Not 100% sure myself. 

We can add a polyfill to the list and add a note to the docs.

taion:
I think for truncating floats, with the current behavior around signs, `Math.trunc` looks better than `parseInt`, when ES6 functionality is available.

For Vimeo, it's actually not a string, so I think numeric comparison should be fine, but I'm not sure and I hadn't checked, because I'm not using Vimeo myself.

Issue #905, opened 2018-04-17
[Accessibility] Multiple Issues making player invalid for WCAG 2.0 AA compliance


fleps:
Updated and more detailed issues reported on https://github.com/sampotts/plyr/issues/905#issuecomment-389909727

----
Original comment:

I'm working on a project that is being converted to meet WCAG 2.0 AA level compliance with the help of a specialized company to do an extensive QA on the project and report all issues found. 
The team found some issue on the Plyr plugin to a point that we will probably need to ditch it and develop our custom player on top of the HTML5 video tag (unless I use the non-minified version and try to fix myself)

I don't expect this issues be fixed soon but I'm reporting them here in order to help, because many accessibility requirements are really hard to understand and we as developer tend to miss-understand some details, so this experience of having a consultant specialist working with us is being really important to learn a few things.

Here are the issues found (PS: I don't know the exact version of the plugin, the min JS doesn't have the version, but for sure is before v3.)

## Slider
Semantic information about the timeline and volume slider control is communicated visually but the current value or the min/max value is not available to all screen reader users.

**Modification Recommended**
- The input ``type="range"`` and ``<progress>`` elements are not fully supported by assistive technology. Use the ``role="slider"``. 
- As appropriate, use WAI-ARIA states and properties to communicate additional semantic information. 
  - Provide a slider label using ``aria-labelledby`` or ``aria-label``. If ``aria-labelledby`` references two or more element ids, set ``tabindex="-1"`` on the elements being referenced.
  - Define the current value using ``aria-valuenow``
  - Set the min/max values using ``aria-valuemax`` and ``aria-valuemin``

## Keyboard Access and Visibility
1 - The video Play, Mute/unmute and full screen button SVG elements are keyboard accessible by default when using Internet Explorer. This may confuse keyboard and screen reader users, as there are two-tab stops for each button.

**Modification Recommended**
- Set ``focusable="false"`` on the ``<svg>`` element.

2 - The volume slider has no visible focus indicator. Mouse users get visible cues that a page element is actionable (mouse pointer change, rollover color change, etc.) that are not available to keyboard-only users.

**Modification Recommended**
- Make sure that there is a clear visual indicator when an object receives focus and when the mouse pointer moves over the object. For example, you could add an outline or change of background color.

## Findable Added Content
[...]

PR #1327, opened 2019-01-30
Update <progress> bar role=progressbar

robertkraig:
aXe reports this as an accessibility issue.

image

aXe is the recommended workflow to test against accessibility by google
https://www.deque.com/blog/google-selects-deques-axe-chrome-devtools/

### Checklist

### DIFF
--- a/src/js/controls.js
+++ b/src/js/controls.js
@@ -347,7 +347,7 @@ const controls = {
                     min: 0,
                     max: 100,
                     value: 0,
-                    role: 'presentation',
+                    role: 'progressbar',
                     'aria-hidden': true,
                 },
                 attributes,


sampotts:
I think if we're going to make this change we can remove the role altogether can't we? Isn't `progressbar` the default for a `<progress>` element? The reason I added `presentation` was to stop updates being announced. See https://github.com/sampotts/plyr/issues/905

robertkraig:
@sampotts I'm working on a project that requires axe validation and that error shows up. Sometimes browsers dont' support `<progress>` so they treat all unknown elements like `<div>` so having the role defined is valid. But notice that axe reports the role being used improperly.

PR #1341, opened 2019-02-13
chore: Move raven-js to devDependencies

taion:
N/A

Move `raven-js` from `dependencies` to `devDependencies`, as it's only used in the demo.

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
         "postcss-custom-properties": "^8.0.9",
         "prettier-eslint": "^8.8.2",
         "prettier-stylelint": "^0.4.2",
+        "raven-js": "^3.27.0",
         "remark-cli": "^6.0.1",
         "remark-validate-links": "^8.0.0",
         "rollup-plugin-babel": "^4.3.2",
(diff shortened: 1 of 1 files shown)


sampotts:
This isn't really the right way to go about it - it's still a dependency of the project as the demo JavaScript imports the module. Ideally what I'd like to do is have a separate package.json (that somehow inherited the root one or perhaps just included the `plyr` package itself) and gulpfile for the demo site.

taion:
Oops, I didn't see the lint thing. We can flag the demo as being a dev thing for linting, though.

I'm not sure I see the difference between e.g. the demo and, say, Gulp. Raven, like Gulp, is a dependency if I'm working on the project itself, but it's not a dependency if I'm installing the `plyr` package via npm.

The demo isn't shipped to npm (https://github.com/sampotts/plyr/blob/80990c98c81fcbd494bc4a14522167d9bd88341a/.npmignore#L1), so the only way that its dependencies would be relevant would be for users who cloned the repo and then installed (all) its dependencies, rather than those installing the package from npm.

Issue #801, opened 2018-03-02
v3 and multiple instances


DanielRuf:
The new version exports only `Plyr` which has no `setup` method.

friday:
You shouldn't upgrade to major software releases and expect them to not break the API.

Changelog etc here: #766

DanielRuf:
I know but it still mentions the `NodeList` example.

DanielRuf:
This does not work:
> const player = new Plyr(document.querySelectorAll('.js-player'));

friday:
From the documentation section I linked to:

"A NodeList or Array of HTMLElement - **the first element will be used**"

This does what you want:

```js
const players = Array.prototype.map.call(document.querySelectorAll('.js-player'), el => new Plyr(el));
```

DanielRuf:
I know. We did it with a loop over the NodeList but the example makes not much sense when the full NodeList is not traversed at all ;-)

friday:
Then don't use NodeList...

sampotts:
Yep setup was removed in favour of a proper class and one to one relationship with element vs instance. It was getting a little messy with allowing whatever in and then spitting out an array with one or many instances back. This way is cleaner and as you've found out if a developer wants to setup multiple players it's as easy as...

```javascript
const players = Array.from(document.querySelectorAll('.js-player')).map(player => new Plyr(player));
```

The example is just to show a `NodeList` is _accepted_ as a value. It doesn't necessarily make it the _best_ for any scenario. It's really because I added support for passing jQuery objects which are always an array of elements as I'm sure you know.

DanielRuf:
> I'll update the docs to make this a little clearer.

Thanks.

PR #1343, opened 2019-02-14
Remove NodeList example

electerious:
I've removed the NodeList example because it's confusing. Only the first element of a NodeList will be used and this is usually not what you want. I've instead replaced it with a `querySelector` and map example.

I know that there was a note, pointing out that the first element will be used, but be honest: Everybody is just scanning the examples for code and doing copy paste :D

https://github.com/sampotts/plyr/issues/801

### Checklist

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -229,7 +229,6 @@ You can specify a range of arguments for the constructor to use:
 
 -   A CSS string selector that's compatible with [`querySelector`](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector)
 -   A [`HTMLElement`](https://developer.mozilla.org/en/docs/Web/API/HTMLElement)
--   A [`NodeList`](https://developer.mozilla.org/en-US/docs/Web/API/NodeList)
 -   A [jQuery](https://jquery.com) object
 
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -248,13 +247,15 @@ Passing a [HTMLElement](https://developer.mozilla.org/en/docs/Web/API/HTMLElemen
...
 ```
 
-Passing a [NodeList](https://developer.mozilla.org/en-US/docs/Web/API/NodeList) (see note below):
+```javascript
+const player = new Plyr(document.querySelector('#player'));
+```
 
 ```javascript
-const player = new Plyr(document.querySelectorAll('.js-player'));
+const instances = [...document.querySelectorAll('.player')].map((elem) => new Plyr(elem))
```
sampotts: I would just remove this completely. Also if we're removing from the docs. We should also remove from the source. The static method can be used to setup multiple players.

electerious: Initialising a page with multiple videos is a common use-case. I guess an example would be helpful.

Removing it from the source would be a good next step.

sampotts: You're correct. That's why it's already covered in the docs:
https://github.com/sampotts/plyr/#setting-up-multiple-players

Removing it should be part of this PR to complete it.

sampotts:
Cheers 👍

Issue #1345, opened 2019-02-15
Keyboard shortcuts do not trigger early-listeners


gurupras:
I use the early-listeners extensively for synchronizing playback. The keyboard shortcuts don't seem to be calling my listeners at all.

### Expected behaviour
Keyboard shortcuts should trigger the early-listeners and use their return value to decide whether to invoke the default handler.

### Actual behaviour
Keyboard shortcuts bypass the early-listeners. Strangely, spacebar alone seems to work under certain circumstances.

### Steps to reproduce
```
function logEvent (msg) {
  console.log(msg)
}

const plyr = new Plyr('video', {
  listeners: {
    play () {
      const msg = `Triggered play: playing=${!plyr.playing}`
      logEvent(msg)
      return true
    },
    seek () {
      const msg = `Triggered seek`
      logEvent(msg)
      return true
    }
  }
})
```
Now, use keyboard shortcuts to control the video playback. If the early-listeners were being triggered, you should see console logs.

### Environment

- Browser: Chrome
- Version: 72.0.3626.96 (Official Build) (64-bit)
- Operating System: Linux
- Version: Ubuntu 18.04

### Link to where the bug is happening
Here's a [pen](https://codepen.io/gurupras/pen/ErdGLZ) describing the issue.

gkralik:
Any chance of merging the related PR #1346. We have hit that issue and the PR seems like it would resolve this.

PR #1346, opened 2019-02-15
Fixed keyboard shortcuts not triggering early-listeners

gurupras:
#1345 

Replaced direct calls with `proxy()` to ensure that default handlers are called first.

Here's a [pen](https://codepen.io/gurupras/pen/pGxGBd) with the fix

### Checklist


src/js/listeners.js
```diff
@@ -99,52 +102,56 @@ class Listeners {
                 case 75:
                     // Space and K key
                     if (!repeat) {
-                        player.togglePlay();
+                        this.proxy(event, player.togglePlay, 'play')
```
sampotts: This should toggle play

gurupras: Sorry, could you elaborate? I don't quite understand what I should do about that line.
Should the `'play'` become `'togglePlay'`?

gurupras: @sampotts, let me know how to fix this when you have some time and I'll fix it so it can be merged in :)

gurupras:
@sampotts, let me know how to fix this when you have some time and I'll fix it so it can be merged in :)

src/js/listeners.js
sampotts: Sorry for mega (!!) late response but I think it should be `togglePlay`?

src/js/listeners.js
```diff
@@ -99,52 +102,56 @@ class Listeners {
...
                 case 37:
                     // Arrow back
-                    player.rewind();
+                    this.proxy(event, player.rewind, 'rewind')
                     break;
 
                 case 70:
                     // F key
-                    player.fullscreen.toggle();
+                    this.proxy(event, () => player.fullscreen.toggle(), 'fullscreen')
```
sampotts: I think you can pass `player.fullscreen.toggle` instead of `() => player.fullscreen.toggle()` ?

src/js/listeners.js
```diff
@@ -99,52 +102,56 @@ class Listeners {
...
                     break;
 
                 case 67:
                     // C key
                     if (!repeat) {
-                        player.toggleCaptions();
+                        this.proxy(event, () => player.toggleCaptions(), 'captions')
```
sampotts: As above

src/js/listeners.js
```diff
@@ -166,7 +173,7 @@ class Listeners {
             // Escape is handle natively when in full screen
             // So we only need to worry about non native
             if (code === 27 && !player.fullscreen.usingNative && player.fullscreen.active) {
-                player.fullscreen.toggle();
+                this.proxy(event, () => player.fullscreen.toggle(), 'fullscreen')
```
sampotts: As above

gurupras:
Will update the code. Thanks for pointing stuff out.

sampotts:
> Will update the code. Thanks for pointing stuff out..

Cheers mate 👍

src/js/listeners.js
gurupras: I can do this, but I don't think the 'pause' event is triggered anywhere else in the code.

https://github.com/sampotts/plyr/blob/977a8393f8a581738650e9472d232d8498d07b11/src/js/listeners.js#L539-L544

Clicking the play/pause buttons only triggers the `'play'` event.

Issue #853, opened 2018-04-01
Plyr 3.0.11 - e.player.elements is null (Firefox Mobile 59, Android 7)


mhluska:
I'm seeing various issues on mobile browsers through Sentry:

https://sentry.io/share/issue/7b6900e51a894fe3b8ab15da365a660e/
https://sentry.io/share/issue/b6ca63233ecc467483aef85643f14fc4/
https://sentry.io/share/issue/0c51b6f3b3fc44b9814d207b36f0987d/

I'm also seeing something on Mobile Safari 11.2.6:

https://sentry.io/share/issue/4c2423d6dcdc40f8a64c6af5379540e3/
https://sentry.io/share/issue/ce2d3d21ac764fda8f75c21fed13ece6/

My Plyr version is 3.0.11.

I don't have these devices myself so I haven't been able to reproduce these issues yet. Are these mobile browsers supposed to be supported?

sampotts:
Ah this looks like a regression. I've not seen ant of these come through on the demo site's Sentry but I'll get these fixed up.

mhluska:
@sampotts thanks for taking a look. To give more detail on the Chrome case, what's happening is I call `.destroy()` on the plyr instance but it doesn't seem to get fully destroyed. This callback seems to not get cleaned up after destroy:

```
          utils.on(this.player.media, 'progress playing', function (event) {
                return ui.updateProgress.call(_this3.player, event);
            });
```

paulconlin:
I'm getting this too across the board on desktop Safari, Chrome, Firefox. Agreed that this has something to do with the `.destroy()`. I see the issue most when a user is cycling through videos on our app, and event listeners are trying to fire on "destroyed" plyr instances.

sampotts:
Are you guys using the latest version?

mhluska:
I am on `3.0.11` still.

FlipgridMartin:
On `3.2.0` right now and still seeing it on multiple browsers and operating systems. They all seem to be coming from inside event listeners.

FlipgridMartin:
[...]

PR #1404, opened 2019-04-01
fix: Properly clear all timeouts on destroy

taion:
Fixes https://github.com/sampotts/plyr/issues/853

In `destroy`, clear some timeouts that previously weren't being cleared.

### Checklist

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -1088,11 +1088,13 @@ class Plyr {
         // Stop playback
         this.stop();
 
+        // Clear timeouts
+        clearTimeout(this.timers.loading);
+        clearTimeout(this.timers.controls);
+        clearTimeout(this.timers.resized);
+
         // Provider specific stuff
         if (this.isHTML5) {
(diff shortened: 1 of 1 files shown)


src/js/plyr.js
```diff
@@ -1088,11 +1088,13 @@ class Plyr {
         // Stop playback
         this.stop();
 
+        // Clear timeouts
+        clearTimeout(this.timers.loading);
```
taion: As far as I can tell, the `loading` and `controls` timeouts will always get set even when using e.g. YouTube, so we should drop them unconditionally.

src/js/plyr.js
```diff
@@ -1088,11 +1088,13 @@ class Plyr {
...
+        clearTimeout(this.timers.controls);
```
taion: I expect this to fix at least the case I'm most often seeing of https://github.com/sampotts/plyr/issues/853.

sampotts:
Thanks for this 👍

PR #1407, opened 2019-04-04
fixed setting youtube host for non-https case

freezer278:
I discovered that in some cases youtube video is not playing when it is in iFrame of non https site (e.g. localhost, etc). This pull request fixes this problem.
### Checklist


sampotts:
Hey,

Thanks for your time. A few issues though:
- You should edit the source files not the built files (i.e. those in `/dist`)
- As the PR checklist shows, you shouldn't commit the changes to `/dist` at all
- As the PR checklist shows, you must PR to `develop` 
- You could have used `window.location.protocol` - e.g.
```javascript
const useSSL = window.location.protocol  === 'https:';
```
- It's You*T*ube. This is a common though
- This sounds more like a browser issue than an issue with Plyr itself

sampotts:
In fact your whole function could be simplified:

```javascript
function getHost(config) {
    if (config.noCookie) {
        return 'https://www.youtube-nocookie.com';
    }

    return `${window.location.protocol}//www.youtube.com`;
}
```

Also it still boils down to the fact I'm not even sure this is required. I've _never_ ran into this before.

freezer278:
Yes this way to write this function is much better. 
This problem is hard to reproduce because i faced it when i use plyr inside iframe that is inside another iframe, and i think it is related somehow to CORS restrictions.

freezer278:
Made some testing in supported browsers, it works.

sampotts:
I meant do we even need the change. Do you have an example where it fails? A codepen, jsfiddle, codesandbox or something?

src/js/plugins/youtube.js
```diff
@@ -34,6 +34,14 @@ function assurePlaybackState(play) {
     }
 }
 
+function getYoutubeHost(config) {
```
sampotts: You're already in the context of YouTube here so it can be just called `getHost`.

freezer278: Yes, you are right. Changed this in last commit.

freezer278:
I am not sure that i can provide you reproduction link. This issue can be reproduced when i use plyr in scorm course that can be opened in lms system deployed locally and it is not open source project and i can`t provide you access to it. 
So i don`t know how to create such case in jsbin because they use https.
I know that this case is very rare and hard to reproduce but accepting these changes doesn't exactly hurt the project, it will only make it work in this rare case.

sampotts:
Cheers 👍

PR #1416, opened 2019-04-15
Make menu work in WebComponent

emielbeinema:
### Link to related issue (if applicable)
#1415 

### Summary of proposed changes
When showing the menu, use a `queryselector` on the container element instead of `document.getElementById` and use the `path` of MouseEvents instead of the target to overcome the limitations of the shadowDOM encapsulation.

### Checklist
- [x] Use `develop` as the base branch
- [x] Exclude the gulp build (`/dist` changes) from the PR
- [ ] Test on [supported browsers](https://github.com/sampotts/plyr#browser-support)

Issue #1415, opened 2019-04-15
Plyr doesn't work in shadowDOM


emielbeinema:
### Expected behaviour
Including Plyr in a WebComponent should work just as it does in normal DOM.

### Actual behaviour
Inside a shadowDOM, the menu closes whenever you click on an item.

### Steps to reproduce
1. Go to below example on codesandbox.io
2. Open the menu
3. Click on one of the items

### Environment

- Browser: Google Chrome
- Version: 75
- Operating System: Ubuntu
- Version: 18.10

### Link to where the bug is happening

https://codesandbox.io/embed/zlkl31j05p

sampotts:
Merged your changes apart for the change to the `matches` method as I think it was a mistake 👍

PR #1417, opened 2019-04-16
Make menu work in WebComponent (continued)

emielbeinema:
#1415 

#1416 introduced some browser-compatibility issues because of the use of non-standard extensions. This pull request aims to fix that (tested on recent Chrome, Firefox, Edge)

### Checklist

### DIFF
--- a/src/js/controls.js
+++ b/src/js/controls.js
@@ -1139,8 +1139,9 @@ const controls = {
             show = false;
         } else if (is.event(input)) {
             // If Plyr is in a shadowDOM, the event target is set to the component, instead of the
-            // element in the shadowDOM. The path, however, is complete.
-            const isMenuItem = popup.contains(input.path[0]);
+            // element in the shadowDOM. The path, if available, is complete.
+            const target = is.function(input.composedPath) ? input.composedPath()[0] : input.target;
+            const isMenuItem = popup.contains(target);
 
             // If the click was inside the menu or if the click
             // wasn't the button or menu item and we're trying to
--- a/src/js/utils/elements.js
+++ b/src/js/utils/elements.js
@@ -234,10 +234,10 @@ export function matches(element, selector) {
     }
 
     const matches =
-        prototype.matches ||
-        prototype.webkitMatchesSelector ||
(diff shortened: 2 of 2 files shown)


src/js/controls.js
```diff
@@ -1139,8 +1139,9 @@ const controls = {
             show = false;
         } else if (is.event(input)) {
             // If Plyr is in a shadowDOM, the event target is set to the component, instead of the
-            // element in the shadowDOM. The path, however, is complete.
-            const isMenuItem = popup.contains(input.path[0]);
+            // element in the shadowDOM. The path, if available, is complete.
+            const target = typeof input.composedPath === 'function' ? input.composedPath()[0] : input.target;
```
sampotts: You can use `is.function(input.composedPath)` here

emielbeinema: Of course, thanks!

src/js/utils/elements.js
```diff
@@ -234,10 +234,10 @@ export function matches(element, selector) {
...
 
     const matches =
-        prototype.matches ||
-        prototype.webkitMatchesSelector ||
-        prototype.mozMatchesSelector ||
-        prototype.msMatchesSelector ||
+        element.matches ||
+        element.webkitMatchesSelector ||
+        element.mozMatchesSelector ||
+        element.msMatchesSelector ||
```
sampotts: What was this change for?

sampotts:
I've manually merged the first change but I think the second is a mistake.

emielbeinema:
Thanks for merging this request. As for the second change, however, I do still think this change is necessary (and correct!) and I suspect that selecting a menu option in the shadowDOM will now omit deselecting the previously selected menu option.

Some supportive evidence that the current code doesn't quite work is how it is compiled. In [`dist/plyr.js`](https://github.com/sampotts/plyr/blob/master/dist/plyr.js#L118), we simply see:
`var matches = match;`, always selecting the custom implementation (that poses problems for shadowDOM).

If you have another proposed solution, I would be happy to implement that, but I do hope we can get this fixed one way or another.

PR #1418, opened 2019-04-18
Fixed problem with displaying subtitles has been solved

HostMIR:
Link: https://codepen.io/anon/pen/dLmdEO
When using subtitles in m3u8, there was no reflection, nothing happened when the button was pressed
### Checklist

### DIFF
--- a/src/js/captions.js
+++ b/src/js/captions.js
@@ -232,6 +232,9 @@ const captions = {
         if (this.captions.currentTrack !== index) {
             this.captions.currentTrack = index;
             const track = tracks[index];
+
+            track.mode = 'showing';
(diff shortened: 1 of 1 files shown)


HostMIR:
> This will break the custom implementation of captions as now both native and custom captions will show. Did you test?

Checked when adding custom subtitles, they are all displayed and working correctly

sampotts:
You didn't check very well then ... With your change the caption is shown twice. This precisely why we set the mode to `hidden` elsewhere as we render captions using custom styling and a custom container. 

image

I can't merge this as-is. You'll need to find another way to fix the issue.

PR #1453, opened 2019-05-24
youtube multiple small issues

aFarkas:
While working with your nice player we had some issues. In case a player is destroyed while the player is still loading the YT api you get an error at `player.media.getAttribute` because player.media is not defined.

While I was fixing this error I also saw that you have some other issues:

1. If setup is called multiple times during yt API loading you re-embed the script (resulted in 7 API scripts on our app)
2. You override a maybe already existing `onYouTubeReadyCallbacks` by a third party script
3. Player declares an needles global Array `onYouTubeReadyCallbacks` (also not standardized)

I haven't tested the code but you should get the issues.

### DIFF
--- a/src/js/plugins/youtube.js
+++ b/src/js/plugins/youtube.js
@@ -47,6 +47,8 @@ function getHost(config) {
     return undefined;
 }
 
+let onYouTubeReadyCallbacks;
+
 const youtube = {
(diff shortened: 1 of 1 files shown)


sampotts:
Awesome! Thank you for doing this 👍

PR #1516, opened 2019-07-30
Toggle also returns promise

azizhk:
Makes toggle also return a promise or undefined.
I know that pause does not return a promise but am still returning it to make it look consistent in the code. I can revert that if needed.

CC: @gehaktmolen 

### Checklist
- [ ] Use `develop` as the base branch


sampotts:
Cheers 👍

PR #1521, opened 2019-08-06
typescript typings

ondratra:
Hi, I have created and tested Typescript types for the project that were missing. 

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
     "homepage": "https://plyr.io",
     "author": "Sam Potts <sam@potts.es>",
     "main": "dist/plyr.js",
+    "types": "src/js/plyr.d.ts",
     "module": "dist/plyr.min.mjs",
     "jsnext:main": "dist/plyr.min.mjs",
     "browser": "dist/plyr.min.js",
--- a/src/js/plyr.d.ts
+++ b/src/js/plyr.d.ts
@@ -1,0 +1,560 @@
+// Type definitions for plyr 3.5
+// Project: https://plyr.io
+// Definitions by: ondratra <https://github.com/ondratra>
+// TypeScript Version: 3.0
+
+export = Plyr;
+export as namespace Plyr;
+
+
+declare class Plyr {
+    /**
+     * Setup a new instance
+     */
+    static setup(targets: NodeList | HTMLElement | HTMLElement[] | string, options?: Plyr.Options): Plyr[];
+
+    /**
+     * Check for support
+     * @param mediaType
+     * @param provider
+     * @param playsInline Whether the player has the playsinline attribute (only applicable to iOS 10+)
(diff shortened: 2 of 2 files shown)


ondratra:
@sampotts can you check this PR, please?

mats852:
In the meanwhile, if anybody needs the type definitions in their project without rebuilding the package themselves, simply add this file to your src:
https://gist.github.com/mats852/5e3a749e0ac050e2d01bb735b7b27cce

I've wrapped @ondratra 's work in a module definition :)

smnbbrv:
@ondratra is there any reason to use `export = Plyr;` instead of `export default Plyr;`?

The current implementation is very confusing since the plyr documentation uses default export all over again and `export = Plyr;` forces to `import * as Plyr from 'plyr'`.

smnbbrv:
@sampotts could you confirm that default export is a "proper" way to go?

ondratra:
@smnbbrv Some clarity on the topic of `export = Something` vs `export default Something` can be found here https://github.com/microsoft/TypeScript/issues/7185 .

The way I import `Plyr` into my project is not via default object - default doesn't seem to work, that's why the typings reflects that via using `export = Plyr`.
```
import * as Plyr from 'plyr'
import Plyr2 from 'plyr'

console.log(Plyr) // ƒ t(n,i) {...}
console.log(Plyr2) // undefined
```

I guess it might seem confusing because `plyr.js` file uses ES import/export, but generated final package (located in `dist/plyr.js`) uses CommonJS - and typings correspond to this final package.

smnbbrv:
@ondratra hm. It works in my project if I use default export... Could you remember what was the issue? 

My problem is that I received a PR https://github.com/smnbbrv/ngx-plyr/pull/35 which is targeting this and now I'm slightly confused. No worries, I can use `* as Plyr` thing; the question is that it's a shared library and I would love to follow the official documented way to use `Plyr` to avoid furhter typings etc problems.

That's why I'd love to see @sampotts opinion in this conversation.

ondratra:
@smnbbrv The issue is mentioned in my last post - `Plyr2` variable is undefined. Check out this sandbox and compare it's `tsconfig.json` to yours, maybe you have some extra options that make it work for you. https://repl.it/@ondratra/plyrTypingsTest If you remove `allowSyntheticDefaultImports: true` from `tsconfig.json` in sandbox, it doesn't compile over that default import.

Strangely enough, at the time of writing that sandbox imports empty object from plyr and I don't know a reason why. =-O But it should return a function like my previous post suggested.

smnbbrv:
@ondratra thanks for claryfying, and great thanks for your work!

I guess the way to use it as default is still to use `allowSyntheticDefaultImports`. Probably should be mentioned in the docs...

PR #1525, opened 2019-08-07
♿️(controls) change play button aria-label value when its state change

lunika:

The aria-label attribute set on all play buttons does not change
according to the player state. When the video is playing, the aria-label
should change to pause otherwise screen reader will not detect that this
button now can be used to pause the video.

### Checklist

### DIFF
--- a/src/js/ui.js
+++ b/src/js/ui.js
@@ -214,6 +214,7 @@ const ui = {
         // Set state
         Array.from(this.elements.buttons.play || []).forEach(target => {
             Object.assign(target, { pressed: this.playing });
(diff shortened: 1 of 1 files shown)


AndrewGulidov:
Possibly developers would want to use their own aria labels. For example "Play [title] video/audio". It would be great to support this case.

lunika:
> Possibly developers would want to use their own aria labels. For example "Play [title] video/audio". It would be great to support this case.

For this you can use your own translations, here I just use a translation key.

sampotts:
Thanks for this 👍

PR #1552, opened 2019-08-27
Fix ads configuration

SoftCreatR:
The defined publisherId is never being transferred to aniview.com due to a typo (there's no `publisherId` in the current scope). Instead, `AV_CDIM2` is always `null` respectively `undefined`.


wehmoen:
Can confirm this. Applied his fix and now the ads work! Thanks @SoftCreatR

sampotts:
Cheers 👍

PR #1565, opened 2019-09-06
Update defaults.js

lmislm:
update defaults.i18n

omit the default i18n key-value
### Checklist


sampotts:
Would this be translated? It’s short for Picture-In-Picture

lmislm:
> Would this be translated? It’s short for Picture-In-Picture

Thank you for your quick response.When display tooltips controls is true, it may need be translated.

sampotts:
Cheers 👍

PR #1570, opened 2019-09-16
Prevents IE11 with resetOnEnd option set to true to play video again

felipedeboni:
When using resetOnEnd after the video finish it start playing again. To prevent this weird behavior and make all browsers behave the same I've added a ```player.pause();``` call right after the ```player.restart();``` without any side effects.

How to reproduce:
https://codesandbox.io/s/static-z964e

Fixed version:
https://fetbx.csb.app/


sampotts:
Cheers 👍

Issue #1492, opened 2019-07-05
Doc on listeners not clear about return value


antonyoneill:
I found the doc a little misleading for the listeners object. I initially added some analytical tracking to some of the buttons, `rewind`, `fast-forward`, `mute` but each of these stopped working.

For example:
```typescript
plyr = new Plyr("video", {
  listeners: {
    mute: event => {
      MyAnalytics.record('mute');
    },
  }
});
```

Only after I looked in the `listeners.js` file did I realise that my event listener had to return a true value.
https://github.com/sampotts/plyr/blob/7c442c93571fb5d75f96bff710e310d00e1289e9/src/js/listeners.js#L511-L518

Also happy to fix this doc issue, just keeping a track of all of these small things and I'll resolve them in one swoop.

sampotts:
That looks like a bug for sure. Wouldn't you might be better off using events for analytics rather than the event listeners though?

taion:
@sampotts We got bit by the same thing. This logic actually looks wrong, though. It looks like this is trying to mimic the browser behavior of preventing defaults when event handlers return `false`, in which case the check should be `returned !== false`, not just `returned`. Not returning a truthy value from a custom event handler should not block the default.

sampotts:
> Not returning a truthy value from a custom event handler should not block the default.

Yep, for sure. Hence:

> That looks like a bug for sure. 

It's on my TODO list but happy to look at PRs to fix it if you're in a rush.

taion:
We can close this out now.

PR #1579, opened 2019-09-25
fix: Fix handling listener return value

### LINKED ISSUE #1491: How can I effectively track user interaction with captions?
I'd like to track caption usage within the player, and I've tried using the methods that are advertised in the documentation but none of them work well enough for my situation.

### `captionsenabled`, `languagechange`, `captionsdisabled`
These plyr events would be ideal to use, however I find that when my stream is loaded (via HLS) if the stream has a text track then both `captionsenabled` and `languagechange` are emitted automatically without user intervention.

Would it be possible some how to include in the event what caused it? Was it automatic, or was it caused by a user?

### `listeners.captions`
This issue was originally about this not functioning at all. The readme references `config/defaults.js` which contains a reference for `captions` but looking in `listeners.js` I can see the custom handler is not used.

This approach actually only solves my problem when the captions button is used directly. If the settings control is available, and the user enables the caption, or changes the language I wouldn't know, using this method.

I would also need to parse the event to determine which track was being activated next, in the case of enable, or change. Since the custom event handlers fire prior to the plyr default event handler.

---

Any advice or thoughts would be awesome.

taion:
Fixes #1491

https://github.com/sampotts/plyr/issues/1492

Only ignore default listener if custom listener explicitly returns `false`.

### Checklist

I'm not sure if any of the built-in plugins depend on the old behavior. I clicked around in the demo for a bit around things like full-screen, and I didn't notice any odd behavior there.

### DIFF
--- a/src/js/listeners.js
+++ b/src/js/listeners.js
@@ -513,7 +513,7 @@ class Listeners {
         }
 
         // Only call default handler if not prevented in custom handler
-        if (returned && is.function(defaultHandler)) {
+        if (returned !== false && is.function(defaultHandler)) {
             defaultHandler.call(player, event);
         }
     }


sampotts:
Thanks for this 👍

PR #1582, opened 2019-10-01
Preserve viewBox attribute in SVG sprite symbols

### LINKED ISSUE #1306: feature: ability to resize buttons
Maybe I miss something out but it is not possible to resize the buttons. We have a plyr player on our page which is quite small and I want to resize all the buttons down to a width of 10px. In #1287 @sampotts suggest CSS... yeah I tried to resize it but it look just strange afterwards:

resize-css-strange

So it would be nice to do this of some other way, maybe a Plyr option? Or is it possible with CSS, then please describe how, I really tried many options.

bseib:
When generating the SVG sprite (using imagemin and svgstore) the imagemin
step needs to NOT cleanup/remove the viewBox attributes from the
individual svg files.

fixes #1306

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -222,7 +222,11 @@ Object.entries(build.sprite).forEach(([filename, entry]) => {
         gulp
             .src(src)
             .pipe(plumber())
-            .pipe(imagemin())
+            .pipe(imagemin([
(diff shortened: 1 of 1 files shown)


sampotts:
Cheers! 👍

Issue #1614, opened 2019-11-14
iOS does not show mute or volume button


TomS-:
### Expected behaviour
I've seen this reported a few times dating all the way back to the start of the year. This really is a breaking bug. I'm having people complain that they can not hear video and it turns out they were all on iOS. I tried explicitly defining the controls in the JavaScript but it doesn't show. Is Plyr still being maintained? 

### Actual behaviour
The mute / unmute and or volume controls should show. 

### Steps to reproduce
1. Visit https://plyr.io/ it does not show and there is no sound.

### Environment

- Browser: Safari
- Operating System: iOS

### Link to where the bug is happening
https://plyr.io/

TomS-:
https://github.com/sampotts/plyr/issues/1310
https://github.com/sampotts/plyr/issues/1439
https://github.com/sampotts/plyr/issues/656
https://github.com/sampotts/plyr/issues/1457

I'm switching to another player until this is fixed.

lnwu:
@TomS- Which player you are switching to? and can volume controls be controled is ios by javascript?

TomS-:
@Inwu I switched to VideoJS and haven't had any issues.

rtarnaud:
It's a simple CSS rule that prevents the volume control to be displayed. Add:
```css
.plyr--is-ios {
    .plyr__volume {
      display: flex !important;
    }
 }
```

to display it again on iOS. That being said, I am not sure why the author of the library decided that the volume control should be hidden in the first place.

sampotts:
This will be fixed in the new release (in a day or two). It was a simple CSS issue. 

> That being said, I am not sure why the author of the library decided that the volume control should be hidden in the first place.

Volume can't be controlled via JS in iOS, only mute/unmute. See https://developer.apple.com/library/safari/documentation/AudioVideo/Conceptual/Using_HTML5_Audio_Video/Device-SpecificConsiderations/Device-SpecificConsiderations.html

sampotts:
Should be fixed in v.3.5.7 👍

Issue #1457, opened 2019-05-28
Audio control on mobile for muted videos and auto play


LittleHamster:
If a vimeo video is set to autoplay and is muted, there is no way to turn on the sound, since the audio control is hidden on mobile. Is it possible to unmute the video as the user tries to turn up the volume on their device?

LittleHamster:
@friday referring to this issue https://github.com/sampotts/plyr/issues/414 is it possible to force audio control?

friday:
I would ask Sam. Haven't touched this code in quite a while.

If you use Vimeos special background video feature there's no way to upgrade to normal video I think.

LittleHamster:
@friday 
I am using the regular video code not the background one, I just can't seem to get the audio working. I even entered the options in control but still nothing. Hopefully @sampotts can provide a solution

edeis53:
When I autoplay or even just play an encoded video with `muted: true`, there is no control visible to unmute it on iOS Safari. However, on desktop Safari it will show the volume control to unmute.

LittleHamster:
any updates on this?

Issue #1310, opened 2019-01-10
Can't un-mute or change volume on muted HTML5 playsinline video on iPad


duncanwilcox:
### Expected behaviour
Ability to un-mute audio on a muted, playsinline video on iPad

### Actual behaviour
No un-mute button or volume slider is available

### Steps to reproduce
- open https://codepen.io/anon/pen/dwqPzO on iPad device or simulator
- start video

### Environment

- Browser: iOS Safari
- Version: 12
- Operating System: iOS
- Version: 12

### Link to where the bug is happening
https://codepen.io/anon/pen/dwqPzO (this is just the basic HTML5 video test case, with the muted attribute added to the video tag)

romukey:
The same thing is happening when i'm using `3.5.6`. 
I need the un-mute button. 
on this issue https://github.com/sampotts/plyr/issues/656, @sampotts says this issue got fixed in `3.2.4`. (I confirmed that although version `3.2.4` throws an error when the page is server-side rendered on nuxt...)

TomS-:
I'm also having this issue, there is no option to mute or unmute a video on iOS.

PR #1629, opened 2019-12-03
IOS - Fix being unable to unmute auto-played videos

sumanbh:
### Link to related issue
#1614
#1457
#1310

We now always show the Volume button so that autoplayed videos on IOS can be unmuted. The input slider won't show on IOS since it does nothing (as the volume level is controlled at OS level).

### Checklist

### DIFF
--- a/src/js/controls.js
+++ b/src/js/controls.js
@@ -1380,7 +1380,7 @@ const controls = {
                 }
 
                 // Volume range control
-                if (control === 'volume') {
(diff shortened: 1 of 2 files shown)


sampotts:
I can see in the screenshot though that there's an extra gap due to removing the CSS to hide the redundant slider. Is that a mistake?

sampotts:
Cheers for this 👍

PR #1651, opened 2020-01-19
Fixed Plyr container not resizing responsively

shravan2x:
The main change was adding `height: 100%`s in a few places. I also had to update `preview-thumbnails.js` to support the new changes.

### Checklist

I use these fixes in a private app and they work fine. However, I had to change them a little when I copied them over. So this PR needs to be tested to verify that
- it fixes the problem and
- it doesn't inadvertently break something else.

I'd appreciate help testing it since I'm not able to do it myself at the moment.

### DIFF
--- a/src/js/plugins/preview-thumbnails.js
+++ b/src/js/plugins/preview-thumbnails.js
@@ -540,8 +540,8 @@ class PreviewThumbnails {
 
     get thumbContainerHeight() {
         if (this.mouseDown) {
-            // Can't use media.clientHeight - HTML5 video goes big and does black bars above and below
-            return Math.floor(this.player.media.clientWidth / this.thumbAspectRatio);
+            const { height } = fitRatio(this.thumbAspectRatio, { width: this.player.media.clientWidth, height: this.player.media.clientHeight });
+            return height;
         }
 
         return Math.floor(this.player.media.clientWidth / this.thumbAspectRatio / 4);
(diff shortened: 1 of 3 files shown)


AkgunFatih:
just tested it and it works

sampotts:
The only styling I'm concerned about is the `height: 100%` on the parent container. Is that needed?

shravan2x:
@sampotts Thanks for the quick response!

The `height: 100%` on the wrapper seemed to be needed in my testing. Did you find otherwise?

AkgunFatih:
this is my testing. i use fancybox 3 with plyr in some of my pages and this is the results i get
**before**
DeepinScreenshot_select-area_20200124092804
**after**
DeepinScreenshot_select-area_20200124092951

kgnfth:
@sampotts Why is `height: 100%;` removed from https://github.com/sampotts/plyr/blob/master/src/sass/base.scss

sampotts:
It was this pull request:
https://github.com/sampotts/plyr/pull/1997

I'm guessing it needs reverting then?

kgnfth:
@sampotts im not the expert but after they removed it, its is back to this issues

image

shravan2x:
When I initially made this PR, the `height: 100%` was definitely required. From that PR it sounds to me like a Bootstrap issue or some missing code on their end.

sampotts:
Could you try `max-height: 100%` on the `.plyr` element? Perhaps that's a happy medium.

kgnfth:
@sampotts 

image

kgnfth:
@sampotts it should be like this https://github.com/sampotts/plyr/pull/1651#issuecomment-578039322

shravan2x:
@kgnfth Could you try adding `max-height: 100%` on both `.plyr` and `.plyr-video-wrapper`? You could also try adding an `object-fit: contain` on the latter.

kgnfth:
@shravan2x same as above

you can try here https://tumbly.gocloud.fun/test/lolvideoslol

shravan2x:
I think I have a solution that should meet fix both this issue and the other one. Working on another PR now.

shravan2x:
@kgnfth Could you try https://github.com/shravan2x/plyr/commit/c6376b6c2b48b06e5528b2e8231b72a9f1fd4f41? If you aren't able to build Plyr entirely you could try making those changes in devtools directly. It's only a few fixes.

sampotts:
Thanks for working this out and sorry for the hassle 👍

shravan2x:
@sampotts Least I could do after all the work you've put into this amazing MIT licensed repository 🙂 

If that fix works, the PR is https://github.com/sampotts/plyr/pull/2022

Issue #1665, opened 2020-01-30
Focus Trap When Exiting Fullscreen 


justinbansal:
### Expected behaviour
When using the keyboard and exiting fullscreen, the expected behaviour is for you to be able to tab forward to the next element. This ends up being the video wrapper usually.
### Actual behaviour
Upon exiting fullscreen, you are stuck in a focus trap. I can't tab forward. If the video is playing, it is stuck here but if video stops, then it will tab forward but instead of going to the wrapper element next, it goes to the play button icon. From this point a tab forward goes to the play/pause button and through all the controls again. It just keeps cycling through this.

### Steps to reproduce
Tab through player (mine is within a Flickity slider), hit enter on fullscreen, hit escape to exit, then try to tab forward.

### Environment

- Browser: Chrome
- Version:
- Operating System: 
- Version:

sampotts:
This should be resolved in v3.5.7. I've used your PR but also fixed the root issue and moved the focus trap logic over to the fullscreen class.

PR #1675, opened 2020-02-07
Trap keyboard focus only when fullscreen

k-jensen:
https://github.com/sampotts/plyr/issues/1665

Include the check 
`!player.fullscreen.active` within the trapFocus function
before trapping keyboard focus.

### Checklist
- [x] Tested on latest Chrome, Firefox, Edge

### DIFF
--- a/src/js/utils/elements.js
+++ b/src/js/utils/elements.js
@@ -257,10 +257,11 @@ export function trapFocus(element = null, toggle = false) {
     const focusable = getElements.call(this, 'button:not(:disabled), input:not(:disabled), [tabindex]');
     const first = focusable[0];
     const last = focusable[focusable.length - 1];
+    const player = this;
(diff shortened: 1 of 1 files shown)


sampotts:
Cheers 👍

sampotts:
Just looking at this again but in theory this shouldn't be needed as when we exit fullscreen we unbind the listener. Doing it this way means we keep re-binding the listener when entering but never unbind it. I'll take a look again.

k-jensen:
I see what you mean, curious as to why the event listener isn't being removed, the type, callback and options seem to match up. Will have another dig at it too.

sampotts:
I’ve moved the logic into the fullscreen class now and just always bind but check within the callback if it’s required. I think it’s a little cleaner anyway.

PR #1676, opened 2020-02-07
Add download attribute to download button

### LINKED ISSUE #1478: when Download button is press it only shows the file
When you try to click the download button it shows the mp3 file to download on a separate sceen than open the file location on where to save it, you have to download the file again or save source as..

Code1110:
Partially fixes #1478. The fix only works for same origin content and browsers that support HTML5.

When clicking the download-button, a new browser window is opened an the content is displayed, instead of showing a save as dialog. To get around this, we can add the `download` attribute to the `<a>` element (see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a).

### Checklist
- [ ] Test on [supported browsers](https://github.com/sampotts/plyr#browser-support) --> Only tested with Firefox and Chrome.

### DIFF
--- a/src/js/controls.js
+++ b/src/js/controls.js
@@ -1580,6 +1580,7 @@ const controls = {
                     element: 'a',
                     href: this.download,
                     target: '_blank',
+                    download: '',
                 });
 
                 const { download } = this.config.urls;


sampotts:
Cheers 👍

PR #1678, opened 2020-02-08
Improve thumbnail size calculations when size is set per css

ydylla:
This allows to set only width or height when setting the thumbnail container size via css. The missing value is calculated base on the thumbnail ratio.

It also fixes the positioning of thumbnail sprites when using a fixed thumbnail size which was set per css.

### Checklist

### DIFF
--- a/src/js/plugins/preview-thumbnails.js
+++ b/src/js/plugins/preview-thumbnails.js
@@ -561,6 +561,11 @@ class PreviewThumbnails {
             return height;
         }
 
+        // If css is used this needs to return the css height for sprites to work (see setImageSizeAndOffset)
+        if (this.sizeSpecifiedInCSS) {
+            return this.elements.thumb.imageContainer.clientHeight;
+        }
+
         return Math.floor(this.player.media.clientWidth / this.thumbAspectRatio / 4);
     }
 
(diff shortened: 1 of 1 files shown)


src/js/plugins/preview-thumbnails.js
```diff
@@ -561,6 +561,11 @@ class PreviewThumbnails {
             return height;
         }
 
+        // If css is used this needs to return the css height for sprites to work (see setImageSizeAndOffset)
+        if (this.sizeSpecifiedInCSS) {
```
sampotts: Where does `sizeSpecifiedInCSS` get set? I can't see it anywhere?

ydylla: It is set in [determineContainerAutoSizing](https://github.com/sampotts/plyr/pull/1678/files/97f8093a8df1fd7462512d43716bcd7601fecb18#diff-4f1c1790f6fc1c2bac82c56401d5e04eL606) which is called in the load function

sampotts: Ahh I missed that! 👍

sampotts:
Cheers 👍

erikt9:
@sampotts @ydylla 
Do you have any guidance regarding how to set the thumbnail size via css?  The clientHeight of the imageContainer always seems be 0 at the time it's checked during initialization, even if I target the container specifically with css.  If I make the rule `!important` the thumbnail container will be the correct size, but will be incorrectly scaled as the clientHeight still evaluated to 0 at initialization.

ydylla:
Hi @erikt9,
it should be enough to set the width or height for `.plyr__preview-thumb__image-container`.   
Here is an example: https://jsfiddle.net/znd0j8gw/

erikt9:
@ydylla Thanks for providing the working demo, I figured out what I was doing wrong (setting `display: none` on `.plyr__controls` until play starts.  The thumbnails are children of that so the clientHeight always ended up being 0).

PR #1701, opened 2020-02-24
Use number instead of string in TS quality definitions

mogzol:
Using strings for the quality doesn't work, plyr expects numbers, so this fixes the definitions.


sampotts:
Cheers 👍

PR #1703, opened 2020-02-25
Completely hide SVG icons to screen readers

LeBenLeBen:
SVG icons should be ignored by screen readers since they have complimentary labels (aria-label or plyr__sr-only). The current « presentation » role simply makes the element behave like a « span » which is incorrect, aria-hidden prevents screen readers from taking care of these elements at all.

### Checklist


sampotts:
Cheers 👍

Issue #1758, opened 2020-04-04
Contextual content in fullscreen


theprojectsomething:
I'm proposing a simple way to extend the player with as little modification to core functionality as possible. This basically involves allowing users to specify an alternative container for fullscreen mode so that other content may also can be overlaid on the player, and remain visible when fullscreen is toggled, without hacking the system.

Examples use-cases include:
- display of video title or other metadata, e.g. tags, donation links, etc.
- alternative access to menu items, such as a searchable captions list (in cases where many hundreds of languages are available)
- custom share dialogs
- integrated playlists with e.g. 'playing next' overlays

The approach should allow bespoke functionality to be created (the kind that has no place in a core library) and integrated without too much thought for anyone else.

I'm writing this having put together a proof-of-concept ready for a PR that I plan using on a decent size project we're about to launch.

helvetica123:
@theprojectsomething hey, I have same problem.I couldnt show my text overlay fullscreen while it seems on smaller screen.I tried below code but it doesnt work for me.If you know how to do it, can you explain me, thanks?

https://stackoverflow.com/questions/52747919/how-to-show-overlay-on-plyr-html5-video-player-when-in-fullscreen/61031057#61031057

theprojectsomething:
Hi @helvetica123 ... I've proposed a solution in the linked PR but I'd say it's a long way from being merged. If you want to try the method out for yourself, you can rebuild the master branch with the changes outlined included (keep in mind it's been created on the develop branch).

Alternatively you can hotlink to a pre-built version of the develop branch, with the changes included, but you wouldn't want to use it for any more than testing (switch out `plyr.js` with whichever build you're after from the dist folder):

https://cdn.jsdelivr.net/gh/theprojectsomething/plyr@a55bf00/dist/plyr.js

sampotts:
Your PR s live in the latest version 👍

theprojectsomething:
> Your PR is live in the latest version 👍

Haha .. I spoke way too soon! Cheers @sampotts

mariusgnicula:
[...]

PR #1759, opened 2020-04-04
Features/fullscreen container

theprojectsomething:
### Related Issue
#1758 

This is a PR to allow for contextual content to be included in fullscreen (or fallback) mode. This means arbitrary elements (extensions to the basic player UI) can be overlaid and remain visible when the player switches to fullscreen.

Example use-cases include:
- display of video title or other metadata (see the included demo)
- alternative access to menu items, such as a searchable captions list (in cases where many hundreds of languages are available)
- custom share dialogs
- integrated playlists with 'playing next' overlays

This approach / PR is just an example of how this feature could work and aims to keep Plyr complexity to a minimum (while enabling some fairly interesting integrations). It utilises a single config option, and does away with the need for injecting bespoke APIs or elements into the player context on a per-project basis. Or trying to mess with what is a pretty slick, but tightly coupled system.

For the user: A new `fullscreen.container` attribute is used to provide a container selector. The container must be an ancestor of the player, otherwise it's ignored. When toggling fullscreen mode, this container is now used in place of the player. Hovering over any children of the container is the same as hovering over the controls. The exception is where the player and the child share a common ancestor (that's not the fullscreen container) ... sounds complex but it's not. You can also gain pretty fine control this way with pointer events.

Under the hood: it adds a `utils/elements/closest` helper method to find the right ancestor. If found this is returned as the fullscreen target in place of the player container. Fullscreen is instantiated slightly earlier in the setup so this container is available for the `listeners.controls` call. In here we add some more 'mouseenter/mouseleave' listeners to any direct descendants of the container, that aren't also ancestors of the player. And that's it. No extra classes, nothing else. There are some style changes to the demo (top margin on the player) but these would be project specific.

Thanks for reading.

### Checklist


src/js/listeners.js
```diff
@@ -814,6 +814,17 @@ class Listeners {
...
 
+        // Also update controls.hover state for any non-player children of fullscreen element (as above)
+        if (elements.fullscreen) {
+            for (let i = 0; i < elements.fullscreen.children.length; i++) {
+              if (!elements.fullscreen.children[i].contains(elements.container)) {
+                this.bind(elements.fullscreen.children[i], 'mouseenter mouseleave', event => {
+                    elements.controls.hover = !player.touch && event.type === 'mouseenter';
+                });
+              }
+            }
```
sampotts, line 825: This could be simplified to:

```javascript
Array.from(elements.fullscreen.children)
    .filter(c => !c.contains(elements.container))
    .forEach(child => {
        this.bind(child, 'mouseenter mouseleave', event => {
            elements.controls.hover = !player.touch && event.type === 'mouseenter';
        });
    });
```

src/js/plyr.js
```diff
@@ -282,6 +283,9 @@ class Plyr {
             });
         }
 
+        // Setup fullscreen
+        this.fullscreen = new Fullscreen(this);
+
```
sampotts, line 288: Why was this moved out of interest?

demo/src/js/demo.js
```diff
@@ -65,6 +65,12 @@ import toggleClass from './toggle-class';
                     // Prevent Vimeo blocking plyr.io demo site
                     referrerPolicy: 'no-referrer',
                 },
+                fullscreen: {
+                    enabled: true,
+                    fallback: true,
+                    iosNative: false,
```
sampotts, line 71: These are the default values and don't need to be specified.

demo/src/js/demo.js
```diff
@@ -65,6 +65,12 @@ import toggleClass from './toggle-class';
...
+                    container: '#container',
```
sampotts, line 72: I'm not sure we need to include this in the demo?

demo/index.html
```diff
@@ -180,6 +180,92 @@ <h1>Pl<span>a</span>y<span>e</span>r</h1>
...
+                            transition: none;
+                        }
+                        .plyr--tags-item small {
+                            display: inline;
+                            margin: 0;
+                        }
+                        .plyr--tags-item::after {
+                            content:  none;
+                        }
+                    </style>
```
sampotts, line 252: Styles should go in the `.scss` files so they are linted, prefixed and formatted correctly. Any colors should also use variables. If you want me to clean this up, I can do that.

package.json
```diff
@@ -31,6 +31,7 @@
     "browserslist": "> 1%",
     "scripts": {
         "build": "gulp build",
+        "serve": "gulp serve",
```
sampotts, line 34: The default task will do this. I'm not sure why you'd want to just serve without having built the project first?

src/js/listeners.js
sampotts: Also, we use 4 space tabs. I'll add a task to format code to make this easier.

theprojectsomething:
Hey @sampotts thanks for coming back so quickly on this - was meant as more of a discussion point than a PR ready to merge ... it's a fairly opinionated solution, figured there'd be some solid feedback. Keen to hear your thoughts, or if there was anything else similar in the pipeline.

Also apologies for the mixup over the `dist` files. Not entirely sure what is meant to happen here, I deleted them from the root folder of the branch, so those shouldn't be in the PR, but assumed `demo/dist` was required for the demo to work so left that in place.

Also the demo is definitely not intended to be merged. It's included in the PR as an example of the feature in use (and for cross device testing). I'll update / respond to the other comments individually. Cheers.

src/js/listeners.js
theprojectsomething: Will fix. Didn't look into your polyfill solution but noticed most of the code seemed to be trying for IE10ish compatibility, heavy prefixing etc.  Figured if I could achieve something in a single loop without es6 I'd do it. TBH I'd usually use `[ ...elements.fullscreen.children ]` :)

src/js/plyr.js
theprojectsomething: The new `elements.fullscreen` is populated duration the `Fullscreen` instantiation, but the listeners for `Controls` were created before this. I checked the module to see if there was any critical reason for the line to be where it was, there didn't seem to be, and so I moved it before `ui.build.call(this)` so I had time to prepare.

theprojectsomething: This is a pretty good example of 'heavily opinionated' (i.e. I didn't have much background on what choices had been made, and so made a decision that worked).

demo/src/js/demo.js
theprojectsomething: I wasn't sure if the config options objects were deep merged, which is why these were included. Again the demo is only there for demoing the feature, can safely be dropped from any merge.

demo/src/js/demo.js
theprojectsomething: Again just demoing the feature .. can safely be dropped from any merge

demo/index.html
theprojectsomething: Sorry, should have made it clear that these are purely for show. Not intended to be included in the merge. They're an example of what someone might do with contextual content. Should I just remove the demo entirely for clarity?

package.json
theprojectsomething: Apologies - definitely ignorance at play there. Haven't used gulp in many, many years. Tried the npm build script but it just recreated the dist and demo folder for me, no file serving. Couldn't get `gulp serve` (or any other gulp commands) working as I don't have gulp installed globally. This worked, and figured it would be useful for anyone else jumping in without any background as I did. Similar to whoever added the gitpod instructions (which I couldn't make head or tail of, seems like an awesome product but kept rebuilding the entire demo on every change .. very slowly).

theprojectsomething:
Ok so I've reverted the demo in the branch which should resolve most of the comments, plus converted the listener logic. Very happy to discuss anything else.

Cheeers!

New demo with fullscreen overlays: https://codepen.io/theprojectsomething/full/bGdyJmv

LittleHamster:
@sampotts is this going to be in the new release soon?

AlGantori:
@theprojectsomething Hello, thank you for this PR.  I was able to id my plyr wrapping div and pass it to plyr configuration.
- I am able to show buttons and a clock.
- Is this overlaying over plyr basically limited to elements needing to exist within this parent container?  
- Perhaps you can explain with a hand sketch what you implemented in this PR. (Promise I won't laugh at your hand drawing skills)
- While i have been longing for this contextual overlaying which I have use for, my clock really does not belong logically to the plyr player component html/ts code.  Even though, at the moment the clock lives there.

What is puzzling me is that I am able to pop a page over the plyr even when it's  in fullscreen and even prior to discovering your "contextual content in fullscreen" and this PR.   You can verify this popup at this stackblitz sample reproduction however, the pits, is that I can't seem to make it appear in my app **when plyr is in fullscreen**.

https://stackblitz.com/edit/ionic-4aduqf?file=pages%2Fhome%2FPlyrPlayerComp.ts

In the above example you will see that plyr is kick into fullscreen without user intervention (no error in the console). you can then click the "Show Dialog" button already overlaid over plyr and it will pop a dialog over plyr.  (everything running sweet).

Here is the fullscreen related issue and my attempts to resolving this
https://github.com/sampotts/plyr/issues/1800 

Perhaps you can shed some lights.  Thank you.

theprojectsomething:
Hi @AlGantori. What you've got working in your demo uses the fallback fullscreen mode which isn't _true_ fullscreen, it just stretches and fixes the player to the document boundaries. This is most useful with iOS devices which have very tight restrictions around what can take over the screen (but depending on your use-case can suffice anywhere).

This PR is intended primarily for true fullscreen mode (though it works for both). With the default setup, when you hit the fullscreen button the player element itself requests to go into fullscreen. This PR simply gives the developer the option to select an element higher in the hierarchy to be controlled by the fullscreen button.

_in some depth ..._
The actual browser fullscreen api is an interesting beast. There are a lot of security/phishing concerns around it due to the fact it completely takes over your screen. Because of this iOS doesn't allow it at all (enter fallback mode), and the other browsers only allow it directly after limited user interaction .. and then tend to throw up garish warnings over the content to make sure you know the website is now in control.

With fallback mode you can safely place anything up over the top of the player with z-indexes and positioning. It's looks and works great, except you don't actually control the whole screen - the browser chrome is still visible. For true fullscreen, the browser essentially projects a limited number of elements into a sandboxed view that takes over the whole screen. No more browser.

... If you think of the dom like a family tree with parents and children. The only things that will appear within that 'true' fullscreen view are children of the dom element that requests fullscreen access.

In your example, the idea would be that you choose an element higher in the dom heirarchy that includes both the player and the clock. If you went as far as to select the `<body>` your entire page would appear.

That said, if your user is hitting the fullscreen button on a video, would they expect a clock to also appear if it doesn't logically belong? If they would expect the clock, perhaps the structural logic is slightly off? Of course you could also consider implementing the fullscreen API separately, outside of the player which might better suit business logic where the player isn't the primary concern.

Hope that helps!

AlGantori:
Hi Sam/@TheProjectSomething, I really appreciated your reply which I have read few times, I understand you must be busy with all the requests so I try to disrupt you the least.
- I think my sample at Stackblitz is a bit misleading given that Stackblitz controls the browser real-estate.  Let me explain a bit more.

- My application contains several players in the PlayersView that it hides and shows.
    1. A video player (plyr)
    2. A talking slide player (slides based on slides.js)
    3. A Quiz player
- I also have **"interrupting"** players that can interrupt current player eg. a calendar, timer, etc... notification
- When such a player kicks in I am trying to sequence carefully the events
1. Message user we are interrupting current play
2. Save current context (user is browsing selections, kiosk mode, a player is playing, ...)
3. Pause current player (eg. plyr)
4. Bring interrupting player to foreground
4. Wait for it complete
1. Message user we resuming previous player
1. Unpause last player (eg. plyr)

- My app is doing two kinds of FullScreen requests, one is at the application level via the following code:
image

- However, in the PlayersView when the current player is plyr I invoke **plyr.fullscreen.enter**
 1. I need advice here, perhaps I have nothing to gain from it going FullScreen? 
and I could just stretch it to occupy the full screen.
- I think I am entering TRUE fullscreen either way, but I could still be wrong in what plyr is actually doing.
 1. I know that from when I started using plyr I sized it (width: 90vw;) such that I left a right border    to keep my plyr alternative buttons visible (eg. FullScreen, Play/Pause)  
 1. I am not CSS stretching plyr when I invoke plyr.fullscreen.enter and it will occupy the full screen.
 1. The main issue I am facing is that Plyr does not succeed and gets blocked by the browser.  Perhaps the issue in **library fullscreen.js** which Plyr seems to be using according to my console?
 1. Here is my css/layout

image

image

image


- Yesterday after reading a post about that weird "1 second barrier" on Firefox, I moved around my code invocation to **plyr.fullscreen.ente**r to invoke it even before .play, and I believe it might also apply to Chromium because I swear I seen it succeed at times.   I have the following video footage I captured last night which only proves that my app/plyr is able to chain another video without plyr ever leaving the "fullscreen"  (a desired scenario)
 1. https://youtu.be/GItAXVy_vxs 
 
- The following short clip #2 demos the following:
 1. We start with regular Chrome from GUI non-maximized.
 1. App goes in FullScreen first (Succeeds)
 1. App switches to the PlayersView 
 1. App tells plyr to .fullscreen.enter and .play  (I get the Error)
 1. https://youtu.be/8IrSLCQGn94 

I am puzzled why I keep getting **"Uncaught (in promise) TypeError: fullscreen error"** ? when the app as far as I  can tell is already in FullScreen (what other user gesture is needed, I understand google/chrome is making these rules).

image


- Regarding the Clock no belonging to Plyr layers, I meant that even though I want it over the Plyr I also want it over the other Players, however, currently when I manually kick Plyr in FullScreen my User Messaging component never appears (blocked?)
So that's why I could not conceive of a "component" like the clock becoming visible over plyr
Note that my User Messaging component appears when PLYR is not in fullscreen.

image

image

- I am not sure I understand your paragraph about "fallback mode" 
image

if all I am experiencing is "fallback mode" how come I seem to be controlling the whole screen, no browser is in view, this happens even without starting the chrome browser in --kiosk mode?
If this paragraph was in response to the sample demo running inside stackblitz, then Nevermind.


- Regarding z-indexes are you saying that in the world of html everything can appear on top 
 1. regardless whether it is nested or not in another element
 2. and some other element is not in fullscreen

By simply styling the desired element to be on top with z-index: some stupidly large number?
In other words position: absolute works also in the Z depth axis?

- I am not sure how I can take advantage of that, yet, given that I am using ionic/angular framework ModalController to pop my orange messages above.

- I think I have said enough :)
- If you want to observe the free app live you may visit the following link and a select a clip for playback, you can later open the dev tools/console.
http://mumti.org/?ch=MUMTI&cat=SLOWTV 

Thanks a bunch for you help

theprojectsomething:
Hi @AlGantori I should probably make clear I don't have anything to do with the plyr library, I've just submitted this particular PR. Also I'm not sure this is the best place to be debugging your issue, as it appears to be based on some confusion with the Fullscreen API / CSS and not with the plyr code itself.

That said, and without getting too far into the weeds, a couple of things to note:

- You can't toggle fullscreen programmatically except off the back off a user gesture (e.g. clicking or tapping a button). AFAIK they need to be in the same callstack. This is the error you're seeing.
- Similarly you can't switch between fullscreen elements without a user gesture (in Firefox anyway, not sure about chromium / webkit)
- IMO if you're already in fullscreen, or implementing the API yourself (which you seem to be, for your entire document no less), you've got little to gain by utilising plyr's fullscreen capabilities ... unless you want to leverage the UI (i.e. fullscreen button) for e.g. stretching the video to the screen dimensions. I think you've got the right idea with your comments above.
- fullscreen.js is just a module in plyr, not an external library. Fullscreen API is relatively straightforward to implement once you have a grasp on the ifs and buts.

Also with regards fallback mode, you're right - it was the stackblitz demo, not your implementation. Sorry for the red herring there!

Apologies if that's not super helpful but I probably have to leave it there. Cheers.

Issue #1737, opened 2020-03-20
Bug in volume option


aubreyz:
### Expected behaviour

Volume option should set initial volume, e.g 
volume: 0.9,

### Actual behaviour

Volume is not responsive at all to this option, either in video or audio, e.g in 

### Steps to reproduce

e.g in

const player = Plyr.setup('.js-playerv', {
    controls,//As above at moment all are uncommented i.e on except the Safari specific ones
   volume: 1.0,
    tooltips:     {    controls:       true  },
    captions: {    defaultActive:  true  }
  });

However a direct change after teh fact does work, e.g

const player = new Plyr('audio', {});
window.player = player;
player.volume = 0.75; // Sets volume at 50%

In fact pointed out previously in a somewhat obscure fashion, but not sure why nobody has noticed this before. https://github.com/sampotts/plyr/issues/848

### Environment

- Browser:
- Version:
- Operating System:
- Version:

sampotts:
I think this is caused by the local storage that stores the user preferences. I'll find a way to force the update if it's passed via the API.

DenisSemionov:
+1, same with `captions: { active: true }` on init

mum-never-proud:
**problem for volume**

https://github.com/sampotts/plyr/blob/fcd82088a5cca625fea16fb60a30640468cb114b/src/js/ui.js#L65

I don't think we need a reset we can go with user config

https://github.com/sampotts/plyr/blob/fcd82088a5cca625fea16fb60a30640468cb114b/src/js/plyr.js#L538-L545

when `build` is called in ui.js we are resetting the volume to null which calls the setter method `volume`, the first preference is given to local storage 

**problem for captions**

https://github.com/sampotts/plyr/blob/fcd82088a5cca625fea16fb60a30640468cb114b/src/js/captions.js#L95-L98

we are giving pref to local storage first rather than config which in my opinion is wrong

I can raise a pr in some time if no one is working on it 

any reason we are resetting the values instead of reading from the config?

https://github.com/sampotts/plyr/blob/fcd82088a5cca625fea16fb60a30640468cb114b/src/js/ui.js#L65-L77

sampotts:
If we always read from config, what is the point of storing the user settings? There's definitely some improvement to be made but ignoring the user settings isn't the answer. If the user sets volume low, then you change source and it shoots up to 100%, it's not a great experience.

DenisSemionov:
For us the initial captions and volume options do not work even when we explicitly disable user options storage.

PR #1761, opened 2020-04-04
init options bug

mum-never-proud:
#1737 

this PR helps in prioritizing the config provided by the User, rather than applying values from local storage, the high priority is given to 

- `user-config`
- `local-storage`
- `app's default config`

it also normalizes options such as quality and speed such that if the user forgets to input one of the required fields it uses app's default config for that field

this code has been tested with `HTML5`, `Youtube` and `Vimeo` and works without issues 

### Checklist
**testing in progress**

### DIFF
--- a/src/js/plyr.js
+++ b/src/js/plyr.js
@@ -25,6 +25,7 @@ import { createElement, hasClass, removeElement, replaceElement, toggleClass, wr
 import { off, on, once, triggerEvent, unbindListeners } from './utils/events';
 import is from './utils/is';
 import loadSprite from './utils/load-sprite';
+import normalizeOptions from './utils/normalize-options';
 import { clamp } from './utils/numbers';
 import { cloneDeep, extend } from './utils/objects';
 import { silencePromise } from './utils/promise';
@@ -37,7 +38,11 @@ import { parseUrl } from './utils/urls';
 
 // Plyr instance
 class Plyr {
-    constructor(target, options) {
+    constructor(target, options = {}) {
+        const { storage } = defaults;
+
+        this.storage = new Storage(storage.enabled, storage.key);
+
         this.timers = {};
 
         // State
@@ -65,17 +70,15 @@ class Plyr {
         // Set config
         this.config = extend(
             {},
-            defaults,
             Plyr.defaults,
-            options || {},
-            (() => {
-                try {
-                    return JSON.parse(this.media.getAttribute('data-plyr-config'));
-                } catch (e) {
-                    return {};
-                }
-            })(),
+            defaults,
+            this.storage.get() || {},
+            options,
+            this.dataPlayerConfig
         );
+
+        this.config.speed = normalizeOptions(this.config.speed, defaults.speed);
+        this.config.quality = normalizeOptions(this.config.quality, defaults.quality);
 
         // Elements cache
         this.elements = {
@@ -257,9 +260,6 @@ class Plyr {
         // Create listeners
         this.listeners = new Listeners(this);
 
-        // Setup local storage for user settings
-        this.storage = new Storage(this);
-
         // Store reference
         this.media.plyr = this;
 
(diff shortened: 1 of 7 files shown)


src/js/plyr.js
```diff
@@ -403,6 +403,14 @@ class Plyr {
...
     }
 
+    get dataPlayerConfig() {
+        try {
+        return JSON.parse(this.media.getAttribute('data-plyr-config'));
+        } catch (e) {
+            return {};
+        }
+    }
+
```
sampotts, line 413: Why do we need this as a getter?

src/js/utils/normalize-options.js
```diff
@@ -0,0 +1,18 @@
+export default function(options, defaults) {
+  if (!options) {
+    return defaults;
+  }
+
+  const normalizedOptions = {};
+
+  Object.keys(defaults)
+    .forEach(key => {
+      if (Object.prototype.hasOwnProperty.call(options, key)) {
```
sampotts, line 10: Why not use `Object.keys(options).includes(key)` ?

src/js/utils/normalize-options.js
```diff
@@ -0,0 +1,18 @@
+export default function(options, defaults) {
```
sampotts, line 1: Is this not just merging the two objects? In which case we can just use the `extend` helper in the object utils?

src/js/utils/objects.js
```diff
@@ -28,10 +28,7 @@ export function extend(target = {}, ...sources) {
 
     Object.keys(source).forEach(key => {
         if (is.object(source[key])) {
-            if (!Object.keys(target).includes(key)) {
-                Object.assign(target, { [key]: {} });
-            }
-
```
sampotts, line 34: What's the reason behind this change? You could potentially overwrite an existing value rather than merge.

src/js/utils/normalize-options.js
mum-never-proud: yep, it is! We have two options with multiple properties `quality` and `speed`

the current implementation will break if the initialization is like

```javascript
const player = new Plyr('#player', {
    speed: {
        selected: 2
    }
});
```
it will break here since the user has not provided the options
https://github.com/sampotts/plyr/blob/39558cbabc76df1fc29fbe12d8f6b14ad01bb559/src/js/controls.js#L1057

the above code just makes sure that incorrect options don't break the code, it wasn't breaking previously since we were overriding user's config

src/js/plyr.js
mum-never-proud: my bad we don't need a getter, but the current implementation is unreadable as well, it confuses the editor that the code below this line is not reachable

I guess we can have a method say `parseJSON(jsonString)` to handle this

src/js/utils/normalize-options.js
mum-never-proud: `hasOwnProperty` is faster with fewer properties say 100, I don't think there's gonna be more than 100 keys here (as of now)

src/js/utils/objects.js
mum-never-proud: I felt like extending an object should extend it unconditionally, rather than checking for the existence of keys in loop

src/js/utils/normalize-options.js
sampotts: Readability > a probably very minor speed increase. You'll notice the rest of the code we use that style so let's stick with that. It's the sort of thing browsers will optimise anyway.

src/js/plyr.js
sampotts: Yeah, it could just be a util really.

src/js/utils/objects.js
sampotts: Did you test it? From looking at it,  happens with your change will be that if the `target` already includes `key`, you're going to wipe it to be an empty object. 

i.e. 
```javascript
const obj1 = { a: { b: 2, c: 3 }};
const obj2 = { a: { d: 4, e: 5 }};
const obj3 = extend(obj1, obj2); 
```

You would expect obj3 to be `{ b: 2, c: 3, d: 4, e: 5 }` but with your change it'll be `{ d: 4, e: 5 }` because it wipes out the previous value.

Issue #1791, opened 2020-04-21
Captions: don't preload until necessary


theprojectsomething:
Currently any `TextTrack` elements are being downloaded as soon as they're associated with a player.  It would be great if the behaviour was a little closer to the browser e.g:

-  individual tracks are fetched only when they're needed (ideal), so on metadata or play / scrub
 - alternatively only the active track is fetched initially on load, with others requested as they become active 

VTTs are generally pretty small, so this really isn't a problem when you've only got one player on a page with one track. But if you've got more than one player with more than one track, then you end up with a whole bunch of content you don't need. In my case I'm working with 5-10 players per page with 100+ tracks (translations) per video, so it's going to get a bit connection heavy.

Looking at the codebase, it looks like things are stemming from `captions.update()` where `mode="hidden"` is being set on every track. I'm wondering if there is a reason for this (I'm guessing a safari/IE quirk) or if it's possible to set the `mode="disabled"` instead, and then set the default track to `mode="hidden"` only when it's needed.

Just a thought starter. Happy to put together a PR if there's a concensus. Thanks for reading.

PR #1793, opened 2020-04-23
Preload TextTracks as per default video element

theprojectsomething:
#1791

These changes bring Plyr captions download behaviour in line with that of the default video element in major browsers. Specifically, text tracks now only download as they are required. Previously all text tracks would download when the Plyr instance was instantiated - which becomes an issue when e.g. many translations are available.

For a track to be downloaded it must either be the default track, the active track when captions are toggled on, or selected from the captions menu.

### Checklist

### DIFF
--- a/src/js/captions.js
+++ b/src/js/captions.js
@@ -133,8 +133,12 @@ const captions = {
                     });
 
                     // Turn off native caption rendering to avoid double captions
+                    // Note: mode='hidden' forces a track to download. To ensure every track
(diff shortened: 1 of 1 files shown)


sampotts:
Thanks for doing this 👍

theprojectsomething:
Hey @sampotts I've just created another related feature request that I'd be keen to submit a PR for. Don't want to spend any time if it's not a good idea for core. Are you able to have a quick squiz / comment on #1803 - it has a demo linked.

PR #1811, opened 2020-04-28
allow custom control as element

taylorchu:

We create custom control, but would like to maintain control element out of plyr.
1. check places where control is passed as "element"
2. don't replace innerHTML. that will get rid of the existing event listener maintained externally.

### Checklist


taylorchu:
@sampotts I closed the old PR. this one should be clearer to review.

PR #1824, opened 2020-05-05
Fixing "missing code in detail" for PlyrEvent type

akuma06:
When using typescript and listening for youtube statechange event, it is missing the code property definition inside the event (even though it is provided in the code).
By making events a map of key-value, we can add easily custom event type for specific event name. Since Youtube "statechange" event differs from the basic PlyrEvent, I added a new Event Type "PlyrStateChangeEvent" having a code property corresponding to a YoutubeState enum as defined by the Youtube API documentation.
This pattern follows how addEventListener in the lib.dom.d.ts is defined.

### Checklist

### DIFF
--- a/src/js/plyr.d.ts
+++ b/src/js/plyr.d.ts
@@ -214,25 +214,25 @@ declare class Plyr {
   /**
    * Add an event listener for the specified event.
    */
-  on(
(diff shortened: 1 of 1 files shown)


sampotts (approved):
Great, thanks!

Issue #1242, opened 2018-10-31
Plugin Support


danrossi:
I see some communication regards to plugin support but I see nothing in here regards to this. Is there a way to easily implement loading plugins. External from the player ? 

The plugin loading needs to happen in the constructor before loading happens.

sampotts:
It's something I'd like to build in but haven't spent much time thinking about how it could be structured. Any help would be appreciated. Making the YouTube and Vimeo players into true plugins would be a good start.

danrossi:
from my experience with flowplayer / video.js. This is a very basic 
plugin registry. They are called before a source is set.  It seems to 
work for me.

I haven't looked further because it lacks HLS / Dash. I was looking to 
port my plugins over but it needs alot of work with the player first.

https://github.com/danrossi/plyr/commit/4a0900bf1821191e4342906a5cae61c457367bb9

------ Original Message ------

filips123:
There should also be support for custom themes.

sampotts:
> There should also be support for custom themes.

There already is via CSS.

filips123:
@sampotts Yes but if plugins will be supported, it will be easier to change it.

PR #1840, opened 2020-05-17
Plugins

akuma06:
#1242 
This Pull request adds the plugin support. Youtube, Vimeo and HTML5 are each a provider of their own. Each provider inherits from the PlyrProvider class. Configs specific to each provider are now inside their static config property. It also make Youtube and Vimeo independent from the code base.
This PR has been tested on Chrome, Firefox and Edge. I don't have access to IE 10 and 11, so I need someone to test it thoroughly.
### Checklist

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,30 +1,31 @@
 {
-    "parser": "babel-eslint",
-    "extends": ["airbnb-base", "prettier"],
-    "plugins": ["simple-import-sort", "import"],
-    "env": {
-        "browser": true,
-        "es6": true
-    },
-    "globals": {
-        "Plyr": false,
-        "jQuery": false
-    },
-    "rules": {
-        "import/no-cycle": "warn",
-        "padding-line-between-statements": [
-            "error",
-            {
-                "blankLine": "never",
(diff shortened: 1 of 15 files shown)


akuma06:
I could potentially update this pr to the last commit. However with no feedback from anyone I don't want to work on it for nothing :). If @sampotts wants at some point to implement this, I will do it though.

Hirbod:
@akuma06 it kinda feels like this repo is somehow dead, considering PRs wont get attention since a year..

sampotts:
> @akuma06 it kinda feels like this repo is somehow dead, considering PRs wont get attention since a year..

It's been quieter than usual, yes. Dead, not really. A lot of PRs were merged in December and January. 
https://github.com/sampotts/plyr/pulls?q=is%3Apr+is%3Aclosed

I'm concentrating on a major development for the next version of Plyr which will mean a total codebase change. I can't reveal too much just yet, suffice to say it will be a major change.

Hirbod:
@sampotts thanks for clarification, didn't meant to be rude or anything. Good news so far. Will there be a plugin system? There is indeed need for other providers and it would be lovely to be able to extend this.

PR #1889, opened 2020-06-30
Add us to the "Used by" list

rutheneum-bote:
We're the students' newspaper of the Goethe Grammar School/Rutheneum since 1608 Gera (Germany) and we use Plyr in order to present the students' songs that were submitted to our school's creativity contest.

It would be really nice if you added us to the "Used by..." list.


sampotts:
Awesome! 👍

PR #1974, opened 2020-10-08
Add missing unit to calc in media query

vincentorback:
Added unit to `calc()` to help interpretation.

I've run into some issues interpreting mixed units in `calc()` inside media queries using [CSSO](https://github.com/css/csso) and [postcss-calc](https://github.com/postcss/postcss-calc). postcss-calc ignores mixed units and csso can’t interpret mixed units. I guess this should be an issue I propos to csso, and I will, but I also think that we should add the missing unit here for a more simple calculation :)

Thanks for the best player out there! <3

### Checklist

### DIFF
--- a/src/sass/components/times.scss
+++ b/src/sass/components/times.scss
@@ -14,7 +14,7 @@
     margin-right: $plyr-control-spacing;
   }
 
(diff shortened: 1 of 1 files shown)


sampotts (approved):
Thanks 👍

PR #1977, opened 2020-10-12
Prevent immediate hiding of controls on mobile when using rewind and fast forward

trafium:
This PR continues #1226 which concerned preventing immediately hiding of controls on mobile after seek, but this time it's about using rewind and fast forward GUI buttons.

Current behaviour: when a touch device user taps the rewind/fast forward button, the controls disappear immediately. Similar behaviour with using seek bar was previously resolved by #1226.

Solution: use the same mechanism implemented by #1226 which is to assign player's lastSeekTime to prevent controls hiding for 2 seconds.

### Checklist

- [x] Test on supported browsers

### DIFF
--- a/src/js/listeners.js
+++ b/src/js/listeners.js
@@ -569,10 +569,18 @@ class Listeners {
     this.bind(elements.buttons.restart, 'click', player.restart, 'restart');
 
     // Rewind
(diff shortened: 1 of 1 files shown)


sampotts (approved):
Thanks 👍

PR #1984, opened 2020-10-18
Fix youtube not working when player is inside shadow dom

xansen:
Youtube videos doesn't work when plyr is inside shadow dom (web components is an example of this use case). It happens because youtube creates an instance of its player using element id, and presumably unable to find it.

Proposal is to use element reference instead of id. This way it works in both plain and shadow dom.

### Checklist
- [ ] Test on [supported browsers](https://github.com/sampotts/plyr#browser-support) (tested on Chrome and Safari)


sampotts (approved):
Thanks 👍

PR #1997, opened 2020-10-30
removes unnecessary height

brunoamaral:
### Link to related issue (if applicable)
https://github.com/sampotts/plyr/issues/1723

### Summary of proposed changes

The height setting doesn't work well with Bootstrap, removing it fixes the problem.

### Checklist
- [ ] Use `develop` as the base branch
- [ ] Exclude the gulp build (`/dist` changes) from the PR
- [x] Test on [supported browsers](https://github.com/sampotts/plyr#browser-support)

PR #2022, opened 2020-11-17
Fixed Plyr sizing in a better way

shravan2x:
Should fix the issue in a way that both #1997 and #1651 work.

### Checklist

### DIFF
--- a/src/sass/base.scss
+++ b/src/sass/base.scss
@@ -25,8 +25,8 @@
   audio,
   iframe {
     display: block;
-    height: 100%;
-    width: 100%;
+    max-height: 100%;
+    max-width: 100%;
   }
 
   button {
--- a/src/sass/types/video.scss
+++ b/src/sass/types/video.scss
@@ -13,8 +13,11 @@
 }
 
 .plyr__video-wrapper {
+  align-items: center;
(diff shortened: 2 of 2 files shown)


src/sass/base.scss
```diff
@@ -13,6 +13,7 @@
   font-variant-numeric: tabular-nums; // Force monosace-esque number widths
   font-weight: $plyr-font-weight-regular;
   line-height: $plyr-line-height;
+  max-height: 100%
```
kgnfth, line 16: Hi you forgot semicolon here

kgnfth:
its still the same

image

image


#1651 this pr was just the best

src/sass/base.scss
kgnfth, line 16: also changing max-height to height will fix this
image

to this
image

dalexhd:
In my case, 3.6.3 version was a bit destructive...

Here is a basic explanation:

[gyazo](https://gyazo.com/25b3ff8685c9a0fb7bc144e3b9bfb3c8)

shravan2x:
Yeah I think we might need to add a new option. I'm still working on this, expect to get somewhere by the weekend.

PR #2029, opened 2020-11-26
Add container over volume input range

ayty-adrianomartins:
- Fix bug in Firefox where volume input range stays over the download buttom

Fix bug in Firefox where volume input range stays over the download buttom
FirefoxBug

### Checklist


sampotts:
Thanks for spending the time on this but the bug has already fixed with a CSS change. Please update your CSS.

PR #2049, opened 2020-12-16
Remove unnecessary calc() from media query

naomiaz:
I've ran into issues compiling my styling using the `gulp-group-css-media-queries` plugin:
image

I discovered the error was in `times.scss`, specifically the line where the media query uses `calc()`. Removing `calc()` solved the compiling error. I guess this should also be fixed in the `gulp-group-css-media-queries` plugin, but since this particular media query works just fine without the `calc()`, I went ahead and removed it :)

### Checklist

### DIFF
--- a/src/sass/components/times.scss
+++ b/src/sass/components/times.scss
@@ -14,7 +14,7 @@
     margin-right: $plyr-control-spacing;
   }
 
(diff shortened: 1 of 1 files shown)


sampotts (approved):

sampotts:
Thanks! 👍

PR #2053, opened 2020-12-22
FIX - object values of the providers object must be used

syteknet-core:
To make use of the provider configuration, the objects values must be used.

### Checklist


sampotts (approved):
Cheers 👍

Issue #2064, opened 2021-01-12
Poster image not visible in Safari 14.0.2


oligafner:
### Expected behaviour
Plyr should show poster image on page load if `data-poster` is set.
### Actual behaviour
Plyr shows black canvas instead of poster image even if `data-poster` is set.
### Steps to reproduce
1. Open Safari 14.0.2 on macOS Big Sur 11.1 (20C69)
2. Go to [plyr.io](https://plyr.io)
3. Look at poster-less player
### Environment

- Browser: Safari
- Version: 14.0.2 (16610.3.7.1.9)
- Operating System: macOS Big Sur
- Version: 11.1 (20C69)

### Link to where the bug is happening
[plyr.io/](https://plyr.io/)

sampotts:
Hey, I'm not seeing this on Safari 14.0.3 (16610.4.3.1.4). This is what I see on plyr.io unless it was fixed in another change 🤔 

<img width="851" alt="image" src="https://user-images.githubusercontent.com/719092/106276698-10b25c00-628c-11eb-8be2-d1c27a584c48.png">

I can't test 14.0.2 as I'm on 11.2. Anyone else seeing this issue?

oligafner:
I've tested on an older MacBook Pro 13" with the same software specs:
Browser: Safari
Version: 14.0.2 (16610.3.7.1.9)
Operating System: macOS Big Sur
Version: 11.1 (20C69)

Result: 
![](https://user-images.githubusercontent.com/6641518/106525429-3fc30a80-64e4-11eb-9835-6922507e4468.jpg)

Another test on a MacBook Air late 2010:
Browser: Safari
Version: 13.1.2 (13609.3.5.1.5)
Operating System: macOS High Sierra
Version: 10.13.6 (17G14042)

Result: 
![](https://user-images.githubusercontent.com/6641518/106525828-df809880-64e4-11eb-9bb3-961f9d2a5947.png)

andrxs:
Hey Guys, I have the same problem. If you want to replicate: Disable captions on demo.js

gchtr:
I’m seeing the same result as @oligafner on https://plyr.io.

MacBook Pro (13-inch, 2019)

Browser: Safari
Version: 14.1.1 (15611.2.7.1.6)

Operating System: macOS Catalina
Version: 10.15.7

![grafik](https://user-images.githubusercontent.com/2084481/124719391-5f34e800-df07-11eb-9860-d2e8a6a2cc93.png)

[...]

PR #2065, opened 2021-01-12
Fix invisible poster in Safari

oligafner:
#2064 
Safari needs to have a set `z-index` on `.plyr__video-wrapper` to not hide the poster image, so we inherit `z-index` from `.plyr`with the value of `0`.
### Checklist

### DIFF
--- a/src/sass/types/video.scss
+++ b/src/sass/types/video.scss
@@ -19,6 +19,7 @@
   overflow: hidden;
   position: relative;
   width: 100%;
+  z-index: inherit;
 }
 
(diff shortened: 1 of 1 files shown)


andrxs:
Hi @oligafner:

That solution fixed poster issue, but on full screen captions and video are overlapping.

regiside:
Hey guys, seems like setting the z-index to 0 explicitly instead of "inherit" solves the issue for me without any side effects. If that's the case for you guys maybe we can update this pull request and get it merged?

PR #2069, opened 2021-01-14
Fix to work inside iframes.

ajgagnon:
Right now Plyr fails to load inside iframes because the selectors are not instances of Element (iframes have their own, separate globals). This is an alternative method to check isElement that will work inside iframes. This is battle-tested fallback code used before browsers supported HTMLElement.

### Checklist


sampotts (approved):
Thanks for this!

Issue #2067, opened 2021-01-13
Call to `Storage.setItem` in `loadSprite` does not catch exceptions


osnard:
If for some reason the local storage can not set the cached sprites, no icons are shown anymore. This can i.e. be the case if the "quota" is already exceeded.

The issue occurs here:  https://github.com/sampotts/plyr/blob/e8d883edba3ee87ff5fbef043ffa50a1a4ae391b/src/js/utils/load-sprite.js#L63-L67

This should probably wrapped into a try/catch block: https://developer.mozilla.org/en-US/docs/Web/API/Storage/setItem#exceptions

### Expected behaviour
Even if the sprites can not be cached on the clients local storage, the control icons should be shown

### Actual behaviour
If the sprites can not be cached on the clients local storage, the control icons are not shown

### Steps to reproduce
Fill your clients local storage to the maximum value

### Environment
- Browser: Chrome 87.0.4280.88

PR #2072, opened 2021-01-15
Add try/catch around `Storage.setItem` calls

osnard:
Issue https://github.com/sampotts/plyr/issues/2067

This is recommended by https://developer.mozilla.org/en-US/docs/Web/API/Storage/setItem#exceptions

### DIFF
--- a/src/js/storage.js
+++ b/src/js/storage.js
@@ -70,7 +70,11 @@ class Storage {
     extend(storage, object);
 
     // Update storage
-    window.localStorage.setItem(this.key, JSON.stringify(storage));
+    try {
+        window.localStorage.setItem(this.key, JSON.stringify(storage));
+    } catch (e) {
+        return;
+    }
   };
 }
 
--- a/src/js/utils/load-sprite.js
+++ b/src/js/utils/load-sprite.js
@@ -60,12 +60,16 @@ export default function loadSprite(url, id) {
         }
 
         if (useStorage) {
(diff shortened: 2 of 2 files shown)


src/js/storage.js
```diff
@@ -70,7 +70,11 @@ class Storage {
     extend(storage, object);
 
     // Update storage
-    window.localStorage.setItem(this.key, JSON.stringify(storage));
+    try {
+        window.localStorage.setItem(this.key, JSON.stringify(storage));
+    } catch (e) {
+        return;
```
sampotts, line 76: The return here does nothing. It can just be an empty catch.

src/js/utils/load-sprite.js
```diff
@@ -60,12 +60,16 @@ export default function loadSprite(url, id) {
...
+            try {
+                window.localStorage.setItem(
+                    `${prefix}-${id}`,
+                    JSON.stringify({
+                    content: result,
+                    }),
+                );
+            } catch (e) {
+                return;
+            }
```
sampotts, line 72: This should be using 2 space tabs. Might be worth running prettier. Don't worry too much if you can't do that. 

Again, the return isn't needed in the catch.

sampotts (commented):
A couple of minor things 👍

sampotts:
I've replicated this logic in the code and made the adjustments I mentioned. Thanks 👍🏼

PR #2076, opened 2021-01-17
Added --plyr-video-background for having control over the background …

nepomuc:
…of a video with alpha channel (webm) or a poster image with alpha channel.

### Checklist
- [✅] Use `develop` as the base branch
- [✅] Exclude the gulp build (`/dist` changes) from the PR
- [💔] Test on [supported browsers](https://github.com/sampotts/plyr#browser-support) ==> Couldn't try it on Apple devices, since I don't own any.


src/sass/settings/colors.scss
```diff
@@ -3,6 +3,7 @@
 // ==========================================================================
 
 $plyr-color-main: var(--plyr-color-main, hsl(198, 100%, 50%)) !default;
+$plyr-video-background: var(--plyr-video-background, rgba(0,0,0,1)) !default;
```
sampotts, line 6: Missing spaces between the RGB values - did you run prettier? It should be `rgba(0, 0, 0, 1)`

sampotts (commented):
Thanks for this. Will just use this PR and close the other. PRs go through develop currently.

src/sass/settings/colors.scss
nepomuc: I ran the gulp "build" job and assumed that it was running everything that's neccessary. I didn't run prettier in addition to that.

sampotts: No dramas. I can sort that out. It wasn't obvious in the docs anyhow.

sampotts (approved):

PR #2108, opened 2021-02-11
Fix errors when Plyr destroyed before setTimeout() executes

emilis-ideait:
Fixed errors when Plyr instance is destroyed before constructor setTimeout() functions execute.

### The issue

We see these errors when we create and destroy a Plyr instance very quickly:

plyr-error-stack-captions-update
plyr-error-stack-constructor-set-speed

Also in some cases the video keeps playing in the background.

Some context:

We are building a React app with a slideshow-like interface. Only the visible slide contents are rendered. Some slides contain Plyr instances that are created in `componentDidMount()` and destroyed in `componentWillUnmount()`.

We noticed that when a user navigates a slideshow very quickly (e.g. via keyboard), the audio from a video in one of the slides keeps playing. We also get the errors in the DevTools console (screenshots above).

I just added the checks if the values are truthy, before assigning properties to them.

### DIFF
--- a/src/js/captions.js
+++ b/src/js/captions.js
@@ -154,7 +154,9 @@ const captions = {
     }
 
     // Enable or disable captions based on track length
-    toggleClass(this.elements.container, this.config.classNames.captions.enabled, !is.empty(tracks));
+    if (this.elements) {
+      toggleClass(this.elements.container, this.config.classNames.captions.enabled, !is.empty(tracks));
+    }
 
     // Update available languages in list
     if (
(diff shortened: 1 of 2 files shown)


liesahead:
This would fix our problem as well!

andyl-astardigital:
+1 is there any chance of getting this PR'd?

abtx:
+1 on this, would be good to have this merged, it's blocking us from using `plyr` at the moment

sampotts (approved):
Thanks 👍🏼

PR #2124, opened 2021-02-25
Migrate color formatting to colorette

jorgebucaran:
Hi! Plyr is using Autoprefixer, Postcss, Rollup, all of which depend on Colorette, so I thought I'd clean up the dep tree a bit.

[Colorette](https://github.com/jorgebucaran/colorette) is also [smaller](https://packagephobia.com/result?p=colorette), and up to 2x faster than ansi-colors, which is nice. No quirks, only usage differences.

Performance numbers comparing `colorette@1.2.2` to `ansi-colors@4.1.1`. Node.js support >= 6.

```
colorette × 788,381 ops/sec
ansi-colors × 350,255 ops/sec
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,6 @@
   },
   "devDependencies": {
     "@sampotts/eslint-config": "1.1.5",
-    "ansi-colors": "^4.1.1",
     "autoprefixer": "^10.0.2",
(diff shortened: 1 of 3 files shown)


sampotts (approved):
Makes sense, cheers for doing this 👍🏼

Issue #1623, opened 2019-11-27
Event: error in triggerEvent.call


laukstein:
Noticed **Event: error** in `triggerEvent.call` on Chrome 65, Linux.
https://github.com/sampotts/plyr/blob/74e3990604698b1f2af3ed51e65b0ca820c1d624/src/js/listeners.js#L499
![grabilla em8812](https://user-images.githubusercontent.com/183966/69709685-89f17480-1106-11ea-961b-19a7666c318d.png)

hasan-wajahat:
Yes Plyr library throws multiple unhandled errors which can be seen on error reporting apps. I have also opened an issue for this as the Plyr floods our error reporting channel.

laukstein:
@hasan-wajahat, what is the issue number you open for it?

hasan-wajahat:
@laukstein The issues I opened was #1581

laukstein:
![grabilla em3900](https://user-images.githubusercontent.com/183966/91472332-f240d280-e89f-11ea-9e41-0fe008a862f0.png)

sacki5:
I still have the same problem. Version 3.6.2
<img width="748" alt="Skärmavbild 2020-09-02 kl  16 15 16" src="https://user-images.githubusercontent.com/25466424/91994968-847d2680-ed37-11ea-85c0-d23f21162e8f.png">

laukstein:
Still same issue on v3.6.3

Issue #1581, opened 2019-10-01
Plyr keeps throwing error without information in error management tools


hasan-wajahat:
### Expected behaviour
I have latest version of Plyr.js on my web application and am using bugsnag to monitor errors. Bugsnag is constantly reporting errors thrown by Plyr but I am not given any further details.
The specific piece of code where the error is generated is shown below, which is basically the function from listener.js

![Screen Shot 2019-10-01 at 4 49 10 PM](https://user-images.githubusercontent.com/10537472/65959465-7eb5fb80-e46b-11e9-89c7-642ee8c849fe.png)

### Actual behaviour
At least some information about the error should be present so I can better understand it.
### Steps to reproduce
I am not sure how to reproduce on any sort of local env.
### Environment

- Browser: Latest versions of Chrome and Safari
- Operating System: Windows, MacOS
- Version: Any version

### Console errors (if any)
No that's exactly what I would like.

hasan-wajahat:
This issue was fixed by https://github.com/sampotts/plyr/pull/1739

PR #2128, opened 2021-03-04
Use prefix for custom event names

chemerisuk:
https://github.com/sampotts/plyr/issues/1581
https://github.com/sampotts/plyr/issues/1623

At present function `triggerEvent` dispatches `CustomEvent` with name in `type` argument. In some cases such events conflict with system ones because use the same name. As a result folks could have weird error records in their reporting tools.

For instance the [`error` event](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror) used to collect all information about errors on page. This event has defined structure and additional fields like `source`, `lineno` etc. But `CustomEvent` with the same name could be dispatched at https://github.com/sampotts/plyr/blob/master/src/js/listeners.js#L515. As the result depended event listeners could fail and display weird results.

The change in PR adds prefix to all dispatched `CustomEvent`s to avoid any naming conflicts with existing events. I'm not sure about potential consequences of this so let me know if any.

### DIFF
--- a/src/js/utils/events.js
+++ b/src/js/utils/events.js
@@ -86,9 +86,14 @@ export function triggerEvent(element, type = '', bubbles = false, detail = {}) {
   if (!is.element(element) || is.empty(type)) {
     return;
   }
+  // error event should not bubble and break error tracking tools
+  // https://developer.mozilla.org/en-US/docs/Web/API/Element/error_event
+  if (type === 'error') {
+    bubbles = false;
+  }
 
   // Create and dispatch the event
-  const event = new CustomEvent('plyr:' + type, {
+  const event = new CustomEvent(type, {
     bubbles,
     detail: { ...detail, plyr: this },
   });
--- a/src/js/utils/events.js
+++ b/src/js/utils/events.js
@@ -88,7 +88,7 @@ export function triggerEvent(element, type = '', bubbles = false, detail = {}) {
   }
 
   // Create and dispatch the event
-  const event = new CustomEvent(type, {
+  const event = new CustomEvent('plyr:' + type, {
     bubbles,
     detail: { ...detail, plyr: this },
   });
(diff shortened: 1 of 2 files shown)


sampotts:
I'd like to see this configurable and opt-in so it's not a breaking change ideally.

chemerisuk:
Hello @sampotts. On the current project we decided to use patched version of `plyr.js` and yes, many errors we had before gone away. Existing custom event dispatching approach is the cause of such errors.

After making the change I've also discovered a bug that produces errors with message `0`. The cause fixed in https://github.com/sampotts/plyr/pull/2128/commits/ddf8332b0ebbd6b458c62eb75d2088eeed8eeeca.

> I'd like to see this configurable (extra property in the config) and opt-in so it's not a breaking change ideally

Keep in mind that other people have errors because of the current custom event names. I'd suggest to force using a unique prefix for event names like other projects do. This way you also avoid potential conflicts with other JS projects.

I understand this is a breaking change and it's better to publish a major release. To improve backward compatibility we can update implementation of `on()`. This method could append event name prefix by default so existing code that uses `on()` will work without any change. What do you think?

sampotts:
I'm happy for this to wait for the next major release if that's the case but I can't guarantee when that would be. I wouldn't bump it to version 4 simply for this change that wouldn't be important for the majority of users. I've used Plyr in several production projects without any issues around event names. Many of them are specific to HTML5 video and audio elements anyway.

chemerisuk:
Hi @sampotts, couple of updates related to this PR. It turns out custom event prefix breaks logic that handles videos in `<iframe>`. As a result Youtube/Video players don't work properly. So I have to rollback the change.

To solve issue with error tracking tools I've just made `error` event not to bubble, so it won't be catched by global handler. BTW according to documentation https://developer.mozilla.org/en-US/docs/Web/API/Element/error_event this event should not bubble. In result we don't see redundant errors in tracking tools and Youtube/Video players work properly.

What do you think about this change?

hasan-wajahat:
Hey @sampotts after the latest changes in this PR there is no longer a prefix added to the custom event. Meaning this should not be a breaking change. Would you please have a look again and see if this can be merged now in version 3.

PR #2191, opened 2021-04-29
Added iPad Os detection to isIos

Benny739:
The new iPad Os is not recognized by the isIos function. this leads to several problems, e.g. fullscreen doesn't work in PWAs on iPad Os.


sampotts (approved):
Thanks! 👍🏼

sparkktv:
This still is broken and I've been testing it. Using plyr 3.6.12 and latest version iPadOS 15.2 and iOS 15.2. And the two are linked somehow with this.
If you have iosnative set to true then iPad full screen DOES NOT WORK. But if you set iosnative to false then full screen on iPad works but then it DOES NOT WORK on iOS (iPhone). It goes full screen but shows the browser window still. It seemed to work fine on iOS/iPadOS 14 with iosnative set to true but now on iOS/iPadOS 15 it's busted again. 

And for my site, we need iosnative set to true, hoping this can be fixed...
It also occurs on the demo site for Plyr, where iosnative must not be set to true because on my iPhone when pressing fullscreen it keeps the browser window open on the top and bottom.

sparkktv:
@sampotts is there a way to look into this because this change seems to have broken in iOS/iPadOS 15+. I'm forced to use Plyr 3.6.7 because it's the last version that iOSnative actually works on both iPhone and iPad on iOS 15+. 
If you have iosnative set to true then iPad full screen DOES NOT WORK. But if you set iosnative to false then full screen on iPad works but then it DOES NOT WORK on iOS (iPhone). It goes full screen but shows the browser window still. It seemed to work fine on iOS/iPadOS 14 with iosnative set to true but now on iOS/iPadOS 15 it's busted again.
It's works without issues however on Plyr 3.6.7 and under before this change was made.

sampotts:
I'll take a look. Thanks for the explanation 👍🏼

sparkktv:
@sampotts Any update on this?

PR #2204, opened 2021-05-10
change fullscreen element to player rootnode

ghost:
Using 'plyr' in a web-component breaks the full-screen feature as the web-component's shadow-root will carry the fullscreenElement instead of the fullscreenElement of document.

### DIFF
--- a/src/js/fullscreen.js
+++ b/src/js/fullscreen.js
@@ -124,7 +124,7 @@ class Fullscreen {
       return hasClass(this.target, this.player.config.classNames.fullscreen.fallback);
     }
 
-    const element = !this.prefix ? document.fullscreenElement : document[`${this.prefix}${this.property}Element`];
+    const element = !this.prefix ? this.target.getRootNode().fullscreenElement : this.target.getRootNode()[`${this.prefix}${this.property}Element`];
 
     return element && element.shadowRoot ? element === this.target.getRootNode().host : element === this.target;
   }


src/js/fullscreen.js
```diff
@@ -124,7 +124,7 @@ class Fullscreen {
       return hasClass(this.target, this.player.config.classNames.fullscreen.fallback);
     }
 
-    const element = !this.prefix ? document.fullscreenElement : document[`${this.prefix}${this.property}Element`];
+    const element = !this.prefix ? this.target.getRootNode().fullscreenElement : this.target.getRootNode()[`${this.prefix}${this.property}Element`];
```
sampotts, line 127: Should `getRootNode()` not be wrapped in a check for `element.shadowRoot` as below?

wdfvdhouten: I am not sure I understand what you mean, but I can explain what the code will do:

when requestFullscreen is called on an element, it will set itself as the fullscreenElement of the parent root.
"Document" in any normal circumstance, a ShadowRoot in case of a web-component.

I was using plyr wrapped in a web component. When entering full-screen, the fullscreen button was not marked as active, because it was looking for the fullscreenElement of the document. (Also pressing F would not exit, it would just trigger requestFullscreen again)

I've made the code change in a private repository for the company I work for, and have made a built of my updated code that we currently use until plyr approves of this patch, and then we would follow the regular release cycle again.

Our fork can be tested here: https://nouacasa-preapproval.financebrokers.ro/
(As you can see, the component sits inside a web component, and therefor the fullscreenElement will not be set on the 'document' node.)

wdfvdhouten: So with the original code, the compare would always be made against the document's fullscreenElement, which can never be the plyr control if it is in the shadowRoot of a web component.

The change just swaps out the reference element to get the fullscreenElement from.

All the things that should be wrapped were already wrapped. Nothing to do additionally.

sampotts: Gotcha, thanks for explaining! 👍🏼

sampotts (approved):

Issue #2230, opened 2021-06-01
SASS Deprecation for division


merlosy:
### Expected behaviour

The sass division notation should be migrated to the new style

### Actual behaviour

While running our application, a deprecation warning appear for plyr regarding Sass division
```sh
DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div(100, 16)

More info and automated migrator: https://sass-lang.com/d/slash-div

   ╷
27 │ $embed-padding: ((100 / 16) * 9);
   │                   ^^^^^^^^
   ╵
    node_modules\plyr\src\sass\types\video.scss 27:19  @import
    node_modules\plyr\src\sass\plyr.scss 44:9          @import
    libs\shared\varda\src\varda.scss 1:9               root stylesheet

WARNING: math.div() will only support number arguments in a future release.
Use list.slash() instead for a slash separator.

   ╷
52 │   margin: -(math.div($plyr-range-track-height, 2)) 0 0;
   │             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   ╵
    node_modules\plyr\src\sass\plugins\ads.scss 52:13  @import
    node_modules\plyr\src\sass\plyr.scss 48:9          @import
    libs\shared\varda\src\varda.scss 1:9               root stylesheet
```

### Steps to reproduce

Use
`@import 'plyr/src/sass/plyr.scss';`
in your main application scss file

### Environment

- Browser: Chrome
- Version: 91
- Operating System: Windows
- Version: 10

### Link to where the bug is happening

Unfortunately it's an internal project

mokipedia:
this is a breaking change, as math.div was added only recently and does not exist on sass prior to v1.32 (i think). Projects using the old node-sass are out of luck there, too.
I would suggest a support wrapper like the angular/component project did:
`// Private polyfill for the `math.div` function from Sass to be used until we can update the
// minimum required Sass version to 1.34.0 or above.
// TODO(crisbeto): replace with `math.div` eventually.
@function private-div($a, $b) {
  @if (meta.function-exists('div', 'math')) {
    @return math.div($a, $b);
  }
  @else {
    @return $a / $b;
  }
}

// Private polyfill for the `list.slash` function from Sass to be used until we can update the
// minimum required Sass version to 1.34.0 or above.
// TODO(crisbeto): replace with `list.slash` eventually.
@function private-slash($a, $b) {
[...]

PR #2234, opened 2021-06-05
Using / for division is deprecated and will be removed in Dart Sass 2.0.0

le0pard:

### DIFF
--- a/src/sass/types/video.scss
+++ b/src/sass/types/video.scss
@@ -21,7 +21,7 @@
 }
 
 // Default to 16:9 ratio but this is set by JavaScript based on config
-$embed-padding: ((100 / 16) * 9);
+$embed-padding: (math.div(100, 16) * 9);
 
 .plyr__video-embed,
 .plyr__video-wrapper--fixed-ratio {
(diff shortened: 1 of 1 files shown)


src/sass/types/video.scss
```diff
@@ -47,7 +49,7 @@ $embed-padding: ((100 / 16) * 9);
 // For Vimeo, if the full custom UI is supported
 .plyr--full-ui .plyr__video-embed > .plyr__video-embed__container {
   $height: 240;
-  $offset: to-percentage(($height - $embed-padding) / ($height / 50));
+  $offset: to-percentage(math.div(math.div($height, $embed-padding), math.div($height, 50)));
```
rrelmy, line 52: It seems the subtraction was also replaced with a division.

Shouldn't it look like
``` scss
  $offset: to-percentage(math.div($height - $embed-padding, math.div($height, 50)));
```
_Not tested!_

le0pard: @rrelmy thanks. Fixed by https://github.com/sampotts/plyr/pull/2234/commits/d7ceaddfce8195b2d07dc89960f076980d4b45df

jan-paulus:
Any update on this?

antonioandrade:
What's the status here? What's stopping this PR from being merged @sampotts ?

antonioandrade:
There seem to be concerns about folks using older versions of node-sass described here https://github.com/sampotts/plyr/issues/2230#issuecomment-879841048

sampotts:
I'll look at the logic suggested in https://github.com/sampotts/plyr/issues/2230#issuecomment-879841048

PR #2256, opened 2021-06-29
Added preview thumbnails setter & fixed cannot set property 'hover' of null

Benny739:
When using custom buttons the hover callback can lead to an error. Especially with youtube "elements.controls" is often null in the beginning. When hovering over one of the custom buttons this leads to 
```
Cannot set property 'hover' of null
```

I also added a setter for preview thumbnails. When working with shaka player for example there is currently no way to refresh the preview thumbnails. 

I also fixed the source types and added the missing previewThumbnails property.


sampotts (approved):
Thanks! 👍🏼

PR #2259, opened 2021-07-02
Let icons load in iframes without a location

ajgagnon:
N/A

Currently, icons will fail to load if the player is placed inside an iframe without a window.location. This adds an additional check to use `window.top.location` if the window does not have a location.


sampotts (approved):
Thanks 👍🏼

Issue #2168, opened 2021-04-12
Captions not working


redd29:
I am using HLS.js to play videos and attaching plyr to the element of the HLS.js, which is working fine. But when I try to load captions from a different url, it doesn't work. (PS. I am using ReactJS)

### Expected behaviour
Subtitles should load and work as expected

### Actual behaviour
Subtitles don't load and word most of the time, but load very occasionally.

The code looks something like
```
if (Hls.isSupported() && streamUrl.includes("m3u8")) {
      const hls = new Hls({ enableWorker: true });

      hls.loadSource(streamUrl);
      hls.on(Hls.Events.MANIFEST_PARSED, () => {
        const qualityLevels = hls.levels.map((level) => level.height).reverse();

        options.quality = {
          default: qualityLevels[0],
          options: qualityLevels,
          forced: true,

          onChange: (quality: number) => {
            const t = quality;
            hls.levels.forEach((e, n) => {
              if (e.height === t) {
                hls.currentLevel = n;
              }
            });
          },
        };
        plyr = new Plyr(element!, options);
      });

      hls.attachMedia(element!);
```

The options I am sending to plyr are
```
  {
      i18n: {
        play: "Play (K)",
        pause: "Pause (K)",
        mute: "Mute (M)",
        unmute: "Unmute (M)",
        enterFullscreen: "Enter fullscreen (F)",
        exitFullscreen: "Exit fullscreen (F)",
      },
      captions: { active: true, language: "en-us", update: true },
      autoplay: true,
      tooltips: { controls: true, seek: true },
      seekTime: 5,
      keyboard: { focused: true, global: true },
      fullscreen: { iosNative: false },
      speed: { selected: 1, options: [0.5, 0.75, 1, 1.25, 1.5, 1.75, 2] },
    }
```

This is how I am loading the subtitles 
```
<video
        autoPlay
        playsInline
        controls
        crossOrigin="anonymous"
        style={{
          width: "100%",
          height: "100%",
          display: "block",
          objectFit: "contain",
        }}
      >
        <track
          kind="subtitles"
          label="English"
          src={enSubUrl}
          srcLang="en-us"
          default
        />
      </video>
```

jackguoAtJogg:
This is a duplicate issue of #1959

Issue #1959, opened 2020-09-21
captions problem when reusing player for another video url


WatchSoMuch:
This happens in all browsers and plyr 3.6.2
I load a video into the player and load captions dynamically by adding <track> elements with src using javascript, captions load and everything is good.
But then I load another video into the player and remove track elements and add another track elements related to second video.
plyr UI shows the old selected language is selected and subtitle is on, but no subtitle is being showed up.

codedealer:
Does not work when setting the new track through the source setter either. `player.currentTrack` shows the previous track loaded but in fact when enabling/disabling captions nothing happens.

WatchSoMuch:
I founded a solution for the problem, setting player.currentTrack for first time sets the track mode to hidden, hidden and showing mode means the browser needs to load the track, when reusing the player, it is needed to make the selected track mode to hidden so it shows up.

etienned:
I also have this bug. After reusing the player with captions, all is OK except that the captions are not shown on the video. I need to switch to another language to get them.

I did an example: https://codepen.io/etienned/pen/abNXOyP

Click on the toggle button to reload the video with and without track. After you toggle it two times (no tracks, tracks again) the captions will not appear until you switch the language.

DaiNguyen1902:
@WatchSoMuch 
Could you please show me more detail about your solution in code? 
Thank you very much

DaiNguyen1902:
I found solution for issue.
I set mode of currentTrackNode is 'showing' and it working
```
var trackNode = player.captions.currentTrackNode;
if(!trackNode || !trackNode.activeCues) return;
trackNode.mode = 'showing';
```

Wiggyboy:
@etienned This doesn't work for me. ( First, captions work ). After toggling twice, the captions don't appear (even though the button and everything) switching language, does not help at all.

I am having a similiar issue like this, #1452. I tried @DaiNguyen1902 suggested solution, but doesn't really seem to help. I am not sure even when to run this, but I run it after switching the source.

Basically, I have many buttons which will play different videos, and my function looks as, below. But the fix did not work.
```
function play(dir) {
	url = "file.php?file=" + dir;
	
	plyr.source = {
		type: 'video',
		title: 'hello',
		sources: [
			{
				type: 'video/mp4',
				src: 'blank.mp4',
			},
		],
	}
	plyr.source = {
		type: 'video',
		title: 'hello',
		sources: [
			{
				type: 'video/mp4',
				src: url + 'media.mp4',
[...]

PR #2261, opened 2021-07-03
Hotfix/captions

zexingguo:
https://github.com/sampotts/plyr/issues/1959
https://github.com/sampotts/plyr/issues/2168

The issue is that the caption doesn't get update when switching video sources. For example, the first video source has caption and when switch to second video source that doesn't, and the caption button stays even though there's no caption. Same issue when you switch from no caption video source to video source with caption. 

I don't think my fix is the best fix since I'm not familiar with the code structure; however, this does solve this issue when switching. My suggestion is to fix in the structure when video source is changed the plyr can detect whether video includes a caption or not and automatically update ui control.


sampotts (approved):
Thanks 👍🏼

tractorcow:
Fixes https://github.com/sampotts/plyr/issues/2352

PR #2303, opened 2021-08-13
Fix invalid CSS selector syntax

BjornBrandewallNaviga:
https://jigsaw.w3.org/css-validator/ complains that "The selector :empty can't appear after the pseudo-element selector ::after". Switching their places, however, will validate.

Fix invalid CSS syntax.

### DIFF
--- a/src/sass/plugins/ads.scss
+++ b/src/sass/plugins/ads.scss
@@ -36,7 +36,7 @@
     z-index: 3;
   }
 
-  &::after:empty {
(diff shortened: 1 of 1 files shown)


sampotts:
I think the issue is that `:empty` won't work here anyway as it should be `[data-badge-text]::after` so that the pseudo element is only shown when the attribute is present. I'll merge this and fix though. Thanks 👍🏼

PR #2319, opened 2021-09-13
Fix types - add tagUrl in ads option with custom VAST

AntLevin:
I`ve added fix for type in plyr.d.ts to fix issue with custom VAST tag. Here https://github.com/sampotts/plyr#options it is

> ads { enabled: false, publisherId: '', tagUrl: '' }

but in real code:
```
  interface AdOptions {
    enabled: boolean;
    publisherId: string;
  }
```


sampotts (approved):
Thanks! 👍🏼

Issue #2321, opened 2021-09-16
[BREAKING CHANGE] Vimeo: supports "h" parameter


rentalhost:
Hello!

Vimeo private videos now requires the parameter `h`. So new uploaded videos are broken and cannot be reproduced at Plyr.

https://github.com/vimeo/player.js/commit/96abf1d435ca226710b993b0154464430e3df70c

Full URL will not works, as suggested by this commit, once that `h` parameter is discarded by Plyr.

Basically, a embed vimeo video have this syntax:

```html
<iframe src="https://player.vimeo.com/video/000000000?h=aabbccddeeff" width="640" height="360" frameborder="0" allow="autoplay; fullscreen; picture-in-picture" allowfullscreen></iframe>
```

Currently Plyr discards `h`, that is required now.

rentalhost:
Workaround:

```js
const plyr = new Plyr(this, {
    vimeo: { h: vimeoPrivateKeyHere }
}),
```

Frosch:
We stumbled upon this new behaviour yesterday, too.
For a bit context: Vimeo changed this for _newly updated videos_ only as of now and seemingly did not really document this, so this behaviour might be a bit puzzling. As some videos (older ones) still work without the hash.

I patched plyr to check the given vimeo-src (if it is a url) for the private hash (url you'll get from oEmbed is `[..]/video/{video-ID}/{hash}` and add the parameter to the iframe if needed.
I'll open the PR in a sec.

bestis:
As the issue has pull request that has been merged this seems to be fixed in the code.

Could we get new release with this fixed? :)

sampotts:
Sure, I just need to work out something with the SASS change that was also merged. Then it’ll all go on npm.

defmeta:
> Sure, I just need to work out something with the SASS change that was also merged. Then it’ll all go on npm.

Hi, sorry to bug you, do you think this will get released today? We are using plyr on an event that's live next week...

PR #2322, opened 2021-09-17
Vimeo private videos: check for hash in src and add as a param

Frosch:

https://github.com/sampotts/plyr/issues/2321

As Vimeo now expects a given hash as a url-parameter `?h={hash}` for _newly uploaded private videos_, I added some code for basic checking whether the given src-url contains the private hash, extracting it and adding it as a parameter to the iframe, if needed.
'Basic' as in, you might want to make this more sophisticated: currently I just strip the src-url, but adding a separate option for vimeo-src or something might be preferred. Also this might be put into some checks whether the user has a business-/pro-account, as otherwise he can't have private videos anyway.

I can change this PR accordingly, if you need me to :)
This currently works as a workaround for us, as you have to do nothing different when using plyr (don't have to split the embed-url by yourself etc.)

### DIFF
--- a/src/js/config/defaults.js
+++ b/src/js/config/defaults.js
@@ -398,6 +398,7 @@ const defaults = {
     embed: {
       provider: 'data-plyr-provider',
       id: 'data-plyr-embed-id',
+      hash: 'data-plyr-embed-hash',
     },
   },
 
--- a/src/js/plugins/vimeo.js
+++ b/src/js/plugins/vimeo.js
@@ -80,6 +80,16 @@ const vimeo = {
       });
     }
 
+    // Get the source URL or ID
+    let source = player.media.getAttribute('src');
+
+    // Check wether there is a hash appended to the given url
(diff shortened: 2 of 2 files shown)


sforward:
Hi @Frosch, I ran into the same problem with the private Vimeo videos, and tested your PR in our project. Some things I noticed: your code breaks the player when no source is set, for example: `<div id="player" data-plyr-provider="vimeo" data-plyr-embed-id="12345"></div>`. The used regex doesn't cover the Vimeo embed url like `https://player.vimeo.com/video/{id}?h={hash}`. In addition, we probably need to add an extra attribute to handle the hash, like `data-plyr-embed-hash`.

erikt9:
I've also found for the oembed api to work, the `h` parameter not only needs to be in the query string, it needs to be the first parameter  (yay, great api vimeo 😝).  Otherwise the oembed call to get the thumbnail will generate a 404.

sampotts (approved):
Just a minor tweak. I can amend this though. Cheers! 👍🏼

vkwc:
Is there timeline when it will go to release?

PR #2410, opened 2022-01-27
feat: add MediaMetadata

Hashen110:
Screenshot from 2022-01-27 16-47-40

add [MediaMetadata](https://developer.mozilla.org/en-US/docs/Web/API/MediaMetadata)


Hashen110:
> Screenshot from 2022-01-27 16-47-40

[MediaMetadata
](https://developer.mozilla.org/en-US/docs/Web/API/MediaMetadata)

Hashen110:
[MediaMetadata](https://developer.mozilla.org/en-US/docs/Web/API/MediaMetadata)

Hashen110:
[Customize media notifications and playback controls outside the browser](https://www.youtube.com/watch?v=Ft6diPReUAU&ab_channel=GoogleChromeDevelopers)

package.json
```diff
@@ -28,7 +28,10 @@
   "bugs": {
     "url": "https://github.com/sampotts/plyr/issues"
   },
-  "browserslist": "> 1%",
+  "browserslist": [
+    "cover 100%",
+    "not dead"
+  ],
```
sampotts, line 34: What's the reason behind this change?

src/js/controls.js
```diff
@@ -1745,6 +1745,21 @@ const controls = {
...
+      if ('mediaSession' in navigator) {
+        navigator.mediaSession.metadata = new window.MediaMetadata({
+          title: this.config.mediaMetadata.title,
+          artist: this.config.mediaMetadata.artist,
+          album: this.config.mediaMetadata.album,
+          artwork: this.config.mediaMetadata.artwork,
+        });
+      }
+      // eslint-disable-next-line no-empty
+    } catch (e) {}
```
sampotts, line 1761: Usually, I just do:
```javascript
catch(e) {
  // Do nothing
}
```

sampotts (commented):
Thanks! I just had one question re: the browserslist change but that's it. The other comment is just a nitpick and can be ignored if you want.

package.json
Hashen110: My IDE (webstorm) showed a warning that string is not a valid option for the browserslist.

Hashen110: And the current value for browserslist does not do anything, I guess

Hashen110: Before changing the browserslist

After changing the browserslist

sampotts: Sorry, your IDE has a bug I guess. It's definitely [valid](https://browserslist.dev/?q=PiAxJQ%3D%3D) and is widely used. The CSS property [`transition`](https://caniuse.com/css-transitions) and [`align-items`](https://caniuse.com/mdn-css_properties_align-items_flex_context) no longer need prefixing so your first screenshot (before the change) is correct.

sampotts (commented):
Thanks! I've reverted the `browserslist` change as discussed.

sampotts (approved):

sampotts:
Thanks for the work on this. I've made a few small changes to the API and UI for it but it's mostly as-is! Thanks again. It's in v3.7.0 🎉

sampotts:
Thanks for the work on this. It's included in v3.7.0 🎉

PR #2413, opened 2022-02-02
[DEV-1642] Upgrade dependencies

nskazki:
This patch upgrades the fork's dependencies altogether. This should fix the debendabot alert and keep this project for a while vulnerability free if we are lucky. After this patch is merged, I'll make the main app use the patched version. The doc on how to do the upgrade is here https://www.notion.so/gleamio/Updating-JS-forks-5d39ab013695444297a6ae69b3a947ec.


nskazki:
Sorry about that! That was meant to be a PR for my fork, but I've picked the wrong upstream.

PR #2436, opened 2022-02-28
SASS → Sass

toastal:
Sass is neither an acronym nor initialism. Sass also was never stylized
as SASS such as Less being stylized as both LESS and {less}.

http://www.sassnotsass.com/


sampotts (approved):
Awesome. Thank you! 👍🏼

PR #2441, opened 2022-03-08
Fix 'this' while switching subtitle tracks

zenyr:
When you switch a subtitle track you'll get

`TypeError: Cannot read properties of undefined (reading 'ui')` in `captions.js`'s `captions.setup`,

Because `captions.setup` is being called without proper `this` binding.

image

Bind this while calling captions.setup


sampotts (approved):
Thanks!

PR #2460, opened 2022-04-18
add markers types

Hashen110:


sampotts (approved):
Thanks! This totally slipped my mind. Thankfully the Vidstack player is written in TypeScript so we'll be exporting auto generated types.

PR #2461, opened 2022-04-18
remove deprecated keyCode

Hashen110:


sampotts (approved):
Thanks! I'm wondering if we should remove other `keyCode` references?

Hashen110:
Yes, We should

PR #2463, opened 2022-04-18
remove deprecated keyCode

Hashen110:


sampotts (approved):
Thanks! I'm working on the listeners logic just in case you thought about looking at it.

PR #2466, opened 2022-04-18
fix invalid CSS @charset rule

Hashen110:
- Charset must be wrapped with double quotes
- `@charset` rule must be placed at the very beginning of the stylesheet


sampotts (approved):

sampotts:
Thanks!

PR #2475, opened 2022-04-24
Fix Player Controllers type declaration. Accept string instead of string[]. Just like controller.md example, which accepts string, not string[]

erez111:
…ing[]. Just like controller.md example, which accepts string, not string[]


sampotts:
I guess it should be `string | Array<string> | ...`? `Array<string>` is still valid.

sampotts:
I've added `string` as per my comment. Commit incoming.

PR #2480, opened 2022-04-28
Empty case inside switch

maiobarbero:


sampotts:
I'm failing to see the issue this PR aims to fix? It looks like it will just break logic.

PR #2489, opened 2022-05-20
Added configurable property to elements for re-use

NoirHusky:
Apparently, if you make a Plyr instance with custom elements. **Plyr declares some of the properties in the element as non-configurable**. So, if you want to **destroy that Plyr instance** and instantiate it again, **Javascript complains of redefining a non-configurable property 'pressed'.** This happens only when the custom elements are an object and not a strings btw.

No, issue precisely about this was opened.

One line was added to make 'pressed' property configurable.


sampotts (approved):

Issue #2199, opened 2021-05-03
Async event listener not invoked


guansss:
### Expected behaviour
Using an async function as event listener.

### Actual behaviour
The async function will never be invoked.

### Steps to reproduce

```js
player.on('ready', async () => {
  console.log('ready');
});
```

I looked into the code and realized the reason is that the async function didn't pass the `isFunction()` check. It checks if the given input's constructor is `Function`, then fails because `(async () => {}).constructor === Function` is false.

IMO it's better to use `typeof input === 'function'`, which works for async functions.

### Environment

- Browser: Chrome
- Version: 90.0.4430.93

dy:
Yep. It is useful to unwrap events to async flows, but now it is blocked:
```

await event(player, 'ready')

// never reached
player.currentTime = 0.1

await event(player, 'timeupdate')

player.elements.container.classList.add('plyr--ready')


// wait until event
function event (target, evt) {return new Promise(r => target.on(evt, function fn(){
  target.off(evt, fn),r()
}))}
```

PR #2497, opened 2022-06-10
Make isFunction safer in case of global pollution of Function class

BakoviDagi:
Our company uses a 3rd-party security service that was wrapping the `Function` class in a `Proxy` and that made the comparison fail, so no videos would start

### DIFF
--- a/src/js/utils/is.js
+++ b/src/js/utils/is.js
@@ -9,7 +9,7 @@ const isObject = (input) => getConstructor(input) === Object;
 const isNumber = (input) => getConstructor(input) === Number && !Number.isNaN(input);
 const isString = (input) => getConstructor(input) === String;
 const isBoolean = (input) => getConstructor(input) === Boolean;
-const isFunction = (input) => getConstructor(input) === Function;
+const isFunction = (input) => typeof getConstructor(input) === 'function';
 const isArray = (input) => Array.isArray(input);
 const isWeakMap = (input) => instanceOf(input, WeakMap);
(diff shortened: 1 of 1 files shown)


src/js/utils/is.js
```diff
@@ -9,7 +9,7 @@ const isObject = (input) => getConstructor(input) === Object;
 const isNumber = (input) => getConstructor(input) === Number && !Number.isNaN(input);
 const isString = (input) => getConstructor(input) === String;
 const isBoolean = (input) => getConstructor(input) === Boolean;
-const isFunction = (input) => getConstructor(input) === Function;
+const isFunction = (input) => typeof getConstructor(input) === 'function';
```
lvdh3000, line 12: I don't think your fix would have the intended effect: a constructor is always a function, so eg. `typeof (10).constructor === "function"` is true but `(10).constructor === Function` is false.

The same line of code causes async functions to not be executed https://github.com/sampotts/plyr/issues/2199
because `(async () => {}).constructor === AsyncFunction`.

I think this should work in all cases:
```suggestion
const isFunction = (input) => input instanceof Function;
```
```js
10 instancof Function //false
(() => {}) instancof Function // true
(async () => {}) instancof Function // true

const proxiedFunction = new Proxy(Function, { construct: (target, args) => new target(...args) });
new proxiedFunction() instanceof Function // true
```

sampotts:
I think we can just simplify the check to `typeof input === 'function'` as per lodash's [`isFunction`](https://github.com/lodash/lodash/blob/master/isFunction.js) method?

Issue #2504, opened 2022-07-14
Unlisted Vimeo videos can't be loaded using the data attributes API


fresswolf:
### Expected behaviour
Using the data attributes API one can load Vimeo Videos like this:
`<div id="player" data-plyr-provider="vimeo" data-plyr-embed-id="https://vimeo.com/40648169"></div>`
Unlisted Vimeo videos recently require an additional hash path segment in the URL: https://vimeo.com/729894358/7767dc808d
We should expect that we can use the same API with the newer URLs like this:
`<div id="player" data-plyr-provider="vimeo" data-plyr-embed-id="https://vimeo.com/40648169"></div>`

### Actual behaviour
When loading an unlisted video we get "Sorry. This video does not exist."

### Steps to reproduce
See below codepen to reproduce the bug.

### Link to where the bug is happening
https://codepen.io/fresswolf/pen/poLNEdM

Issue #593, opened 2017-05-29
Playing back private vimeo videos


misha-panyushkin:
- [ ] Issue does not already exist

### Expected behaviour 
Authorization via app credits or playing back anyway if allowed

### Actual behaviour
404 on requesting video via `oembed`

### Environment

- Browser: Chrome
- Version:  Version 58.0.3029.110 (64-bit)
- Operating System: OSX
- Version: Latest

Players affected:
- [ ] HTML5 Video
- [ ] Vimeo

### Steps to reproduce 
- provide private full link to the allowed for embedding vimeo video

Chalderrr:
Any chance this is being looked into or if there is a temporary workaround?

albalooshi:
Hi, We are also interested in this

ctsstc:
To my knowledge private vimeo videos are just `/videoId/secretId`
I also noticed on their Share -> iframe it puts the url as `https://player.vimeo.com/video/videoId?h=secretId`

maxfenton:
It’s not documented, but it _seems_ like this _works_ when `data-plyr-embed-id="[complete vimeo URL including id and hash]"`, but does _not_ work with just `data-plyr-embed-id="[id]/][hash]"` —

PR #2505, opened 2022-07-14
fix: allow unlisted vimeo video urls when loading using data attributes (#2504)

fresswolf:
#2504 

Also see #2322 which fixed the issue, except for the case when data attributes are used.
Also this old ticket #593 can probably be closed with this fix.

This allows us to use URLs of unlisted Vimeo videos in the data-plyr-embed-id data attribute. Currently, data-plyr-embed-id does work with full vimeo URLs like this
`<div id="player" data-plyr-provider="vimeo" data-plyr-embed-id="https://vimeo.com/40648169"></div>`
but it doesn't work when the URL contains an additional hash (unlisted video):
`<div id="player" data-plyr-provider="vimeo" data-plyr-embed-id="https://vimeo.com/729894358/7767dc808d"></div>`

I found an undocumented data attribute called data-plyr-embed-hash which can be used to set the hash separately. But if we allow setting full URLs using the data-plyr-embed-id attribute, I would expect to be able to pass the URL including the hash to this attribute without having to care about the privacy settings of the video myself.

### DIFF
--- a/src/js/plugins/vimeo.js
+++ b/src/js/plugins/vimeo.js
@@ -24,8 +24,10 @@ function parseId(url) {
     return url;
   }
 
-  const regex = /^.*(vimeo.com\/|video\/)(\d+).*/;
-  return url.match(regex) ? RegExp.$2 : url;
+  const regex = /^(.*vimeo.com\/|.*video\/)?(\d+).*/;
+  const found = url.match(regex);
+
+  return !found ? url : found[found.length - 1];
 }
 
 // Try to extract a hash for private videos from the URL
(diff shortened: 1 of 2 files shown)


src/js/plugins/vimeo.js
```diff
@@ -94,6 +97,11 @@ const vimeo = {
       source = player.media.getAttribute(player.config.attributes.embed.id);
       // hash can also be set as attribute on the <div>
       hash = player.media.getAttribute(player.config.attributes.embed.hash);
+      // In case hash is not explicitly set as attribute we try to parse it from the id attribute
+      if (!hash) {
+        hash = parseHash(source);
+        alert(hash);
```
sampotts, line 103: I guess we need to remove the `alert()` here?

PR #2516, opened 2022-08-14
plyr.io used on Thousands of Index Websites

PBhadoo:

* Added https://gitlab.com/GoogleDriveIndex/Google-Drive-Index Project

We've given credits to plyr.io too.
https://gitlab.com/GoogleDriveIndex/Google-Drive-Index#credits


sampotts (approved):
Thanks!

Issue #2323, opened 2021-09-17
plyr is not compatible with most recent dart sass versions


julmot:
### Expected behaviour

plyr v3.6.8 (most recent) can be compiled with dart sass and postcss.

### Actual behaviour

plyr v3.6.8 (most recent) does not compile successfully (throws error).

```
ERROR in ./node_modules/plyr/dist/plyr.css
Module build failed (from ./node_modules/mini-css-extract-plugin/dist/loader.js):
ModuleBuildError: Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
SassError: -7px and -2 are incompatible.
    ╷
483 │   margin-right: calc(-7px - -2);
    │                      ^^^^^^^^^
    ╵
  node_modules\plyr\dist\plyr.css 483:22  root stylesheet
```
(plyr.css needs to be formatted in order to generate this specific error message, the error message is useless when not formatted)

It might be due to the changes of [dart sass v1.40.0](https://github.com/sass/dart-sass/releases/tag/1.40.0).

> calc() expressions will be parsed more thoroughly, and errors will be highlighted where they weren't before. This may break your stylesheets, but only if they were already producing broken CSS.

### Steps to reproduce

simply compile plyr with sass and postcss using most recent versions.

------

Since May I already had to downgrade plyr to v3.6.4 since it wasn't compatible with most recent versions of dart sass and postcss using webpack ([reference](https://github.com/sampotts/plyr/issues/2182#issuecomment-830099098)). Now, with the latest releases of dart sass it's not even compiling v3.6.4 anymore.

julmot:
I can confirm plyr v3.6.4 is not working with [Dart Sass 1.40.0](https://github.com/sass/dart-sass/releases/tag/1.40.0), but works with 1.39.2.

julmot:
For anyone struggling with the same issues. plyr seems to be incompatible with both most recent dart sass (see above) and postcss (see https://github.com/sampotts/plyr/issues/2182#issuecomment-833676995). With webpack it's therefore necessary to add `exclude: /plyr\.css$/` to existing scss and css file loaders. Full config:

```
    { // sass and postcss at the end
        test: /\.(scss|css)$/,
        exclude: /plyr\.css$/, // see above
        use: [
          MiniCssExtractPlugin.loader,
          {
            loader: 'css-loader'
          }, {
            loader: 'postcss-loader',
            options: {
              postcssOptions: {
                config: path.join(__dirname, './postcss.config.js')
[...]

PR #2519, opened 2022-08-16
Fix for calc() in newer Dart Sass versions

ckhicks:

[Issue 2323](https://github.com/sampotts/plyr/issues/2323) 

Several people (including myself) encountered a bug as Dart Sass became more strict with its `calc()` function. This specific line without a `px` unit breaks compilation.


sampotts (approved):
Thanks! Sorry for the delay on this.

PR #2523, opened 2022-08-30
Improve accessibility on control buttons with aria-pressed

emilkarl:
Improve accessibility by adding `aria-pressed` attribute to control buttons and sync with together with the `pressed` attribute.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed


emilkarl:
Ping @sampotts 👯

sampotts (approved):

PR #2530, opened 2022-09-14
navigator.platform is deprecated

stamat:
I'm submitting a little update for browser detection:

1. `navigator.platform` is deprecated - it's prefered to use `navigator.userAgent` https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform
2. No need for the brackets in the RegEx - you can test this on https://regex101.com

And that's about it. Maybe I missed the point, do let me know! ❤️


sampotts (approved):
Thanks!

Issue #2104, opened 2021-02-08
PIP on mobile Safari


ezadoo:
As iOS14 supports Picture-in-Picture on mobile too, I think it would be a good improvement, if the PIP-Button would be displayed on mobile-safari too.

mayeaux:
+1. Yeah I was wondering what was going on and why it wasn't visible. Maybe I can add a PR.

PR #2533, opened 2022-09-21
fix iPhone no PIP bug

happyjake:

https://github.com/sampotts/plyr/issues/2104

iPhone can do PIP too. So exclude iPhone from PIP support  is pretty outdated. Let's remove that limitation.

### DIFF
--- a/src/js/support.js
+++ b/src/js/support.js
@@ -38,10 +38,6 @@ const support = {
   // Picture-in-picture support
   // Safari & Chrome only currently
   pip: (() => {
-    if (browser.isIPhone) {
-      return false;
-    }
-
     // Safari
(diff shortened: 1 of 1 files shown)


sampotts (approved):
Thanks!

sampotts:
Hey, was this change tested? It looks like while picture-in-picture is possible on an iPhone, we can't manually trigger it like on the iPad. I may have to revert this change.

sampotts:
It will throw an error in the console:
> Unhandled Promise Rejection: NotSupportedError: The Picture-in-Picture mode is not supported.

Reverting. Next time please give it a test, even the simulator that's part of Xcode is good enough 👍🏼

PR #2540, opened 2022-10-08
i18n: Make captions autodetect text direction

### LINKED ISSUE #2043: Wrong text direction for right-to-left captions
### Expected behaviour

Subtitle/caption texts should be displayed in the right 'direction', depending on the language.  For instance, English and French captions should be displayed left-to-right (they are); and Persian, Arabic, and Hebrew should be displayed right-to-left (they currently are not).

### Actual behaviour

Captions for right-to-left languages like Persian, Arabic, and Hebrew are displayed left-to-right (incorrect) rather than right-to-left (correct).

### Workaround / Fix

The workaround/fix is trivial: set the `dir="auto"` attribute on the captions container.  I currently apply this workaround using `player.elements.captions.dir = "auto";`, where `player` is a `Plyr` instance.  This should be the default behaviour of Plyr, so that right-to-left captions are displayed correctly out of the box.

ebraminio:
This uses browser mechanism for detecting appropiate direction for text, dir=auto.

Fixes #2043

This uses `dir="auto"` so subtitles written in RTL scripts (Persian, Arabic, Hebrew and etc) will have a better direction using plyr.

Use of dir=auto in such scenarios are very common, for example see this very page source and find `dir=auto` that is used all over the user content to enhance the display of user text that can be in different language even in GitHub. Years ago I added the thing in Gitlab also [here](https://gitlab.com/gitlab-org/gitlab-foss/-/merge_requests/6296) years ago before GitHub's adoption with the improvement and now as plyr was used by someone that needed this fix am adding the the improvement here also.

image

### DIFF
--- a/src/js/captions.js
+++ b/src/js/captions.js
@@ -47,6 +47,7 @@ const captions = {
     // Inject the container
     if (!is.element(this.elements.captions)) {
       this.elements.captions = createElement('div', getAttributesFromSelector(this.config.selectors.captions));
+      this.elements.captions.setAttribute('dir', 'auto');
 
       insertAfter(this.elements.captions, this.elements.wrapper);
     }


sampotts (approved):
Awesome. Thank you!

PR #2548, opened 2022-10-23
fix menu border radius bug

raad-altaie:
fix for css customizing bug `--plyr-menu-radius`


sampotts (approved):
Good catch. Thanks!

PR #2549, opened 2022-10-23
fix tooltip seek render bug

raad-altaie:
`white-space: normal; `

Screen Shot 2022-10-23 at 5 53 48 PM

`white-space: nowrap;`
Screen Shot 2022-10-23 at 5 53 17 PM

### DIFF
--- a/src/sass/components/progress.scss
+++ b/src/sass/components/progress.scss
@@ -27,7 +27,7 @@ $plyr-progress-offset: $plyr-range-thumb-height;
     left: 0;
     max-width: 120px;
     overflow-wrap: break-word;
-    white-space: normal;
+    white-space: nowrap;
   }
 }
 


src/sass/components/progress.scss
```diff
@@ -27,7 +27,7 @@ $plyr-progress-offset: $plyr-range-thumb-height;
     left: 0;
     max-width: 120px;
     overflow-wrap: break-word;
-    white-space: normal;
+    white-space: nowrap;
```
sampotts, line 30: Thanks for the change and sorry for the slow review! We can remove this line completely since the default `white-space` value for `.plyr__tooltip` is `nowrap`.

sampotts:
Thanks for doing the work on this. I've just removed the line as per my comment for now. Cheers!

PR #2567, opened 2022-11-11
Upgrade vulnerable packages along with the rest 

nskazki:
# Changes

This patch upgrades a vulnerable package along with the rest.

Dependabot alert:

https://github.com/Crowd9/plyr/security/dependabot/5

# Testing

There's nothing to test. The only difference in the output is a hundred removed newlines and even more rearranged comments.


nskazki:
Sorry about that. It's a local fork we have been maintaining for a while.

Issue #2490, opened 2022-05-20
toggle playback


d4lv:
the "space"  toggle playback stop working in the last version but only on browser like chrome-brave-opera etc. on firefox is ok is any fix for that??

(when i play it on my site the page is go down and in the demo https://plyr.io/ is just dont do nothing but the K is ok)

fugbixer:
@sampotts 
please add space keyboard shortcut for pause / resume video playing

RaSan147:
@sampotts any update??

WebLeach:
Just use version 3.7.0 for now. 
Tried to fork and re build with older ones, but required packages weren't installing

d4lv:
still this is not fixed on new version..

Rishi-Bidani:
> @sampotts still this is not fixed on new version..
 I am using this as a work around till its fixed.

```
    (document.querySelector(".video__container") as HTMLDivElement).addEventListener("keydown", (e) => {
        if (e.key === " ") {
            e.preventDefault();
            (video.value as any).plyr.togglePlay();
        }
    });
```

PR #2581, opened 2022-12-02
Fixed event key with space

royeden:
https://github.com/sampotts/plyr/issues/2490

The library switched from `event.keyCode` to `event.key` for listeners here:  https://github.com/sampotts/plyr/releases/tag/v3.7.1

But the space bar wasn't working because the `event.key` for spacebar is an empty space string `" "`. See the [mdn reference](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values#whitespace_keys).

Thank you for such an asewome library! :smile:

### DIFF
--- a/src/js/controls.js
+++ b/src/js/controls.js
@@ -404,7 +404,7 @@ const controls = {
       'keydown keyup',
       (event) => {
         // We only care about space and ⬆️ ⬇️️ ➡️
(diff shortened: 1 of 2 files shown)


sampotts (approved):
Thanks for this!

Issue #2210, opened 2021-05-19
Preview Thumbnail's listeners() function never called - causes some issues


mogzol:
### Expected behaviour

The `listeners()` function in `preview-thumbnails.js` should be called, and the listeners within it should be set up.

### Actual behaviour

They aren't. Listeners set up in the `listeners.js` file still happen, so preview thumbnails _mostly_ work, but there are some issues, such as:

- Preview thumbnails aren't hidden on play/seek (using keyboard)
- Scrubbing overlay is not hidden if video is paused and you scrub somewhere, then let go of mouse after video has loaded at current time (due to `previewThumbnails.lastTime` never being set)

### Steps to reproduce

Put a breakpoint or console log inside the `listeners()` function in `preview-thumbnails.js`. Observe that it is never hit.

### Environment

- Plyr version: 3.6.8

### Console errors (if any)

N/A

### Link to where the bug is happening

Can be observed on the demo player at https://plyr.io/

PR #2604, opened 2023-01-19
Call preview-thumbnails `listeners()` function on load

mogzol:

#2210

The `listeners()` function in preview-thumbnails.js was never called, so some functionality was broken. This change calls listeners in the preview-thumbnails load function.


sampotts (approved):
Thanks!

Issue #2624, opened 2023-02-20
Vimeo player mute button doesn't work on mobile phones any more


dowixxx:
I can see the same behaviour on https://plyr.io/#vimeo. On desktop everything works fine, but when i checked on a few different phones "mute" button doesn't work any more (it does not mute the video), even at plyr.io homepage. 

Please fix this!

HandreMelo:
I had the same issue here, but only on IOS.

PR #2628, opened 2023-02-24
fix(issue #2624): Changing Vimeo.js function used to mute video.

HandreMelo:

In the file Vimeo.js, the function responsable to set player muted, uses setVolume(0) but it didn't work on IOS. On Vimeo's API reference, it says "NOTE: Most mobile devices don't support a volume level independent of the system volume. In these cases, this method always returns 1." (https://developer.vimeo.com/player/sdk/reference) , so i changed the function call to setMuted().


sampotts (approved):

sampotts:
Thanks for this! 👍🏼

PR #2679, opened 2023-05-11
Support en2015

rezakho:
Remove Nullish coalescing operator.
Support es2015 (like android 7 blrowser)


NamPNQ:
You need to change gulp build config to support this, instead change source code

Issue #2749, opened 2023-11-14
The progress bar is offset


Wshengs:
### Expected behaviour
The progress bar is working properly.
### Actual behaviour
The progress bar is offset.
![6801699954683_ pic_hd](https://github.com/sampotts/plyr/assets/27401205/36191254-70d3-413a-b53c-f8e2dafb3f42)

### Steps to reproduce

1. Please adjust the width of the page so that the scrollbar appears
2. Scroll the page to the far right and click on the progress bar

### Environment

- Browser: Chrome 
- Version: 114.0.5735.248 (arm64)
- Operating System: MacOs
- Version: 12.0

### Link to where the bug is happening
[https://codepen.io/s-chen/pen/XWOaoda](https://codepen.io/s-chen/pen/XWOaoda)

PR #2750, opened 2023-11-14
fix:progress offset

Wshengs:
#2749 
Minus scrollLeft width


sampotts (approved):
Thanks for this!

PR #2857, opened 2025-06-02
docs: fix Koel's tagline and URL

phanan:

Koel's tagline and URL have changed since the last time :)


sampotts (approved):
Cheers!

PR #2859, opened 2025-08-08
add Mux as an example in the demo

dylanjha:
- [x] add Mux as an example in the demo (after HTML5 video/audio)
- [x] Also added a link to mux.com for folks seeking video hosting
- [x] Added a small line linking to Mux for video hosting in the README
- [x] Removed vidstack callout in the README

A few complications in the demo.js code to call out:

- Since Mux is an HLS source, this required loading hls.js which I am doing from the CDN at the bottom of the demo.html file
- Re-using the media element for an HLS source and then toggling back to a non-hls source did not work cleanly, so in order to do this I needed to add 2 elements on the page, a `#player` and a `#player-hls`
- There's a bit of hacky-code to show/hide the right player (and pausing the underlying media element when hiding
- I found that for Plyr instances of type `"video"`, you could set player.media.hidden = true and that would hide the player
- But for Plyr instances of type `"audio"`, that didn't work, so it required setting `hidden` on the `player.elements.container`

Testing

- Everything seems to be working, I tried loading in different states and toggling between the different options and things are switching as expected

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,6 @@
-| 🎉  | [Plyr is merging into Vidstack](https://github.com/sampotts/plyr/issues/2408) | 🎉  |
-| :-: | :---------------------------------------------------------------------------: | :-- |
-
 Plyr is a simple, lightweight, accessible and customizable HTML5, YouTube and Vimeo media player that supports [_modern_](#browser-support) browsers.
 
-[Checkout the demo](https://plyr.io) - [Donate](#donate) - [Slack](https://bit.ly/plyr--chat)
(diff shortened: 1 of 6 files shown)


sampotts:
Thanks @dylanjha. I'll check this out later today fix up those GitHub actions. I'm guessing we should probably close the related issue about the Vidstack merge and maybe add some info in there?

sampotts (dismissed):

sampotts:
> @sampotts got it! Thanks for the feedback.
> 
> I agree we should discuss this on the Vidstack issue would that be best coming from you, you think?
> 
> Happy to help with any other cleanup

Sure, I'll add an update today and get this change deployed.

PR #2879, opened 2025-12-12
fix: remove unused --plyr-font-smoothing variable and set default smoothing

### LINKED ISSUE #2873: There is no use of the `--plyr-font-smoothing` variable inside CSS source code
Inside `plyr.css` you can't find `--plyr-font-smoothing` getting used at all.

In SCSS code, we can see this:

https://github.com/sampotts/plyr/blob/287ec5a0b256613f9db5d03477c51cad6dd73a45/src/sass/settings/type.scss#L16

and this:

https://github.com/sampotts/plyr/blob/287ec5a0b256613f9db5d03477c51cad6dd73a45/src/sass/lib/mixins.scss#L14-L15

This isn't valid code, because the current default value is a string, so it's always true.

So, the whole thing is useless. What to do, just make it the default and remove variables.

AbdelatefElshafei:
Fixes #2873
- Kept the SCSS variable $plyr-font-smoothing and CSS custom property --plyr-font-smoothing so users can override it.
- Updated the plyr-font-smoothing mixin to respect the $mode argument, ensuring that explicit calls like @include plyr-font-smoothing(true) work correctly.
- Default font smoothing is now true (antialiased for webkit, grayscale for Firefox).
- Ensures font smoothing now applies consistently across all elements.

### DIFF
--- a/src/sass/lib/mixins.scss
+++ b/src/sass/lib/mixins.scss
@@ -11,13 +11,15 @@
 
 // Font smoothing
 // ---------------------------------------
-@mixin plyr-font-smoothing($mode: true) {
-  @if $mode {
+@mixin plyr-font-smoothing($mode: $plyr-font-smoothing) {
+  @if $mode == true {
+    -webkit-font-smoothing: antialiased;
     -moz-osx-font-smoothing: grayscale;
-    -webkit-font-smoothing: $plyr-font-smoothing; // Used the correct Var
+  } @else {
+    -webkit-font-smoothing: auto;
+    -moz-osx-font-smoothing: auto;
   }
 }
-
 // <input type="range"> styling
 // ---------------------------------------
 @mixin plyr-range-track() {
--- a/src/sass/lib/mixins.scss
+++ b/src/sass/lib/mixins.scss
@@ -14,7 +14,7 @@
 @mixin plyr-font-smoothing($mode: true) {
   @if $mode {
     -moz-osx-font-smoothing: grayscale;
-    -webkit-font-smoothing: antialiased;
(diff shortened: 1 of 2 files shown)


AbdelatefElshafei:
> @AbdelatefElshafei Just out of curiosity, could you elaborate on your changes? It looks as if `$plyr-font-smoothing` is still in the code (you said it should have been removed?). Also I don't get why you are ignoring the method argument `$mode` in the mixin `plyr-font-smoothing()`? Thanks for your efforts! Cheers, Ole.
> 
> Edit: To clarify, I meant that in your changed code, the mixin `plyr-font-smoothing()` changes its behavior depending on the setting of the scss variable. If you set `$plyr-font-smoothing` to false and call `plyr-font-smoothing(true)`, font-smoothing will still be disabled, right?

@gardiner Hi Ole,

Thanks a lot for your detailed feedback! You're absolutely right.

I've updated the PR to address your concerns:

The SCSS variable $plyr-font-smoothing is still present as the default, so users can override it via CSS or SCSS.

The mixin plyr-font-smoothing($mode) now respects the $mode argument — calling @include plyr-font-smoothing(true) will enable font smoothing regardless of the default variable.

The default behavior remains consistent, and all elements now apply font smoothing correctly.

This should fix the issue you pointed out about the mixin ignoring the argument.

Thanks again for reviewing — happy to make any further adjustments if needed!

Best,
Abdelatef

AbdelatefElshafei:
> @AbdelatefElshafei Awesome, thank you!
You are Welcome, Thanks For your FeedBack, Ole!

PR #2888, opened 2026-01-22
fix: sync captions to native iOS fullscreen player

### LINKED ISSUE #2732: Captions not displaying in fullscreen when using iosNative: true
### Expected behaviour

Switching to iosNative should preserve captions.

### Actual behaviour

Captions are not visible

### Steps to reproduce

https://codepen.io/awerks/pen/ZEVXddm

### Environment

- Browser: Safari, Chrome
- Version: latest
- Operating System: IOS
- Version: IOS 16.6

mkrn:
## Summary
- Fixes captions not displaying in native iOS fullscreen when using `iosNative: true`
- When entering native fullscreen, sets active caption track mode to `'showing'` so iOS can render captions
- When exiting, restores track mode to `'hidden'` so Plyr can resume controlling caption rendering
- Also fixes a bug where `exit()` was calling `webkitEnterFullscreen()` instead of `webkitExitFullscreen()`

## Test plan
- [x] Test on iOS device with `iosNative: true` configured
- [x] Enable captions and enter fullscreen
- [x] Verify captions are visible in native iOS fullscreen player
- [x] Exit fullscreen and verify Plyr captions still work normally

Fixes #2732

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/js/fullscreen.js
+++ b/src/js/fullscreen.js
@@ -55,6 +55,16 @@ class Fullscreen {
     // Tap focus when in fullscreen
     on.call(this, this.player.elements.container, 'keydown', event => this.trapFocus(event));
 
+    // Handle iOS native fullscreen caption sync
+    if (browser.isIos && this.player.config.fullscreen.iosNative) {
+      on.call(this.player, this.player.media, 'webkitbeginfullscreen', () => {
+        this.syncCaptionsToNative(true);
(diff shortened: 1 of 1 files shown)


mkrn:
Tested at preview URL with iPhone. LGTM

PR #2901, opened 2026-07-08
fix(captions): case-insensitive language matching + keep selected track

gurupras:
Two related caption language-matching fixes:

- Match caption tracks case-insensitively. The wanted language is lowercased, but it was compared against the raw track `srclang`, so a track declared as e.g. `srclang="EN"` never matched a preferred language of `en`. Lowercase both sides in `update()` and `findTrack()` (and guard null srclang).

- Keep the currently selected track when several share a language. `setLanguage` always resolved to the *first* track for a language via `findTrack`, so users with multiple same-language tracks could never switch to (or stay on) the others. Prefer the current track when it still matches the requested language, falling back to `findTrack` otherwise.

### DIFF
--- a/src/js/captions.js
+++ b/src/js/captions.js
@@ -120,7 +120,7 @@ const captions = {
     const tracks = captions.getTracks.call(this, true);
     // Get the wanted language
     const { active, language, meta, currentTrackNode } = this.captions;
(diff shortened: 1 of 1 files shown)


sampotts (approved):

sampotts:
Thanks!

PR #2905, opened 2026-08-13
fix: restore volume when unmuting from zero

### LINKED ISSUE #2904: Mute button does nothing after setting volume to 0
### Expected behaviour
If volume is at 0, toggling mute should restore a usable volume so playback can be heard again — especially when no volume slider is available to raise volume manually.

### Actual behaviour
At volume 0 the mute control looks muted, but toggling it only flips the `muted` flag and leaves volume at 0. With no volume slider (or when the slider isn’t usable), the user has no way to get sound back and is stuck silent.

### Steps to reproduce
1. Open https://plyr.io/
2. Play the sample video
3. Set volume to 0 (or reach a state where volume is 0)
4. Use the mute button / mute control
5. Observe that audio stays silent and there is no way to recover volume when a slider isn’t available

### Environment
- Browser: Chrome
- Version: 151.0.7922.110
- Operating System: Mac
- Version: 26.6.1

### Console errors (if any)
None

### Link to where the bug is happening
https://plyr.io/

hamedniroomand:
### Link to related issue
Fixes #2904 

- Treat volume `0` as muted when toggling mute
- Restore the last audible volume on unmute so users aren’t stuck silent when no volume slider is available


sampotts (approved):
Thanks!

